PR #16, opened 2016-08-04
Case sensitivity cleanup

peterp:
A quick cleanup, something more focused is required. Subsequent pull-requests incoming.


src/components/Button.js
```diff
@@ -0,0 +1,81 @@
...
+  buttonText: {
+    color: '#fff',
+    fontFamily: 'Brandon Text',
+    fontSize: 13,
+    fontWeight: 'bold',
+    marginTop: 15,
+    marginBottom: 17,
+    letterSpacing: 1,
+  },
+});
```
justinvdm: Are these styles moved from elsewhere in the codebase? Just wondering if there needs to be a corresponding deletion.

src/components/Button.js
```diff
@@ -0,0 +1,81 @@
...
+  button: {
+    justifyContent: 'center',
+    alignItems: 'center',
+    backgroundColor: '#f77040',
+    borderRadius: 5,
+    marginLeft: 16,
+    marginRight: 16,
+  },
+  buttonIsDisabled: {
+    backgroundColor: '#dfe5e6',
```
justinvdm: Maybe not for this PR, but wondering if we should alias our colours as constants in some style vars module.

src/components/Button.js
```diff
@@ -0,0 +1,81 @@
...
+  setNativeProps(nativeProps) {
+    this._root.setNativeProps(nativeProps);
+  }
+
+  render() {
+    let button = (
+      <View
+        style={[
+          styles.button,
+          this.props.disabled ? styles.buttonIsDisabled : undefined,
```
justinvdm: [A bit nitpicky](https://shapeshed.com/the-void-of-undefined-in-javascript), but wondering if we could use `null` here instead (provided React will handle it the same way as `undefined`)?

justinvdm:
:+1: looks good :) left questions, but I don't think they should prevent this from landing.

src/components/Button.js
peterp: They were in `./src/StyleSheet.js,` I will eventually be moving everything out of there.

src/components/Button.js
peterp: 👍

src/components/Button.js
peterp: 👍

PR #19, opened 2016-08-05
Refactoring Mentor to Auth

peterp:


src/auth/ActivationContainer.js
```diff
@@ -0,0 +1,29 @@
+import { connect } from 'react-redux';
+import { withRouter } from 'react-router';
```
peterp: This is a wrapper that adds `router` to the context of our containers. So that we're able to transitions to different routes without having to couple `react-router` directly to the component.

src/auth/Login.js
```diff
@@ -0,0 +1,69 @@
+/* eslint-disable react/jsx-no-bind */
```
peterp: I use this component as a base view for both logging in and account activation, since they're essentially the same thing but with different shades of lipstick.

They're customised and differentiated by the container components: `LoginContainer` and `ActivationContainer`

src/auth/actions.js
```diff
@@ -0,0 +1,51 @@
+import base64 from 'base-64';
+import { normalize, Schema, arrayOf } from 'normalizr';
+
+import { getBaseURL, getAuthorizationToken } from 'src/configuration.js';
```
peterp: React doesn't have globals, so I've settled on these... I don't like that the actions are coupled to them though.

src/auth/actions.js
```diff
@@ -0,0 +1,51 @@
...
+  AUTH_LOGIN_REQUEST,
+  AUTH_LOGIN_SUCCESS,
+  AUTH_LOGIN_FAILURE,
+  AUTH_PASSWORD_RESET_REQUEST,
+  AUTH_PASSWORD_RESET_SUCCESS,
+  AUTH_PASSWORD_RESET_FAILURE,
+} from './constants';
+
+
+const profileSchema = new Schema('profile');
```
peterp: Perhaps I need to creates a schemas.js file; they can be more involved than just this simple declaration.

src/auth/actions.js
```diff
@@ -0,0 +1,51 @@
...
+  })
+  .then(response => response.json())
+  .then(json => {
+    const entities = normalize(json.results, arrayOf(profileSchema));
+    if (entities.result.length === 0) {
+      return dispatch({ type: AUTH_LOGIN_FAILURE });
+    }
+
+    // TODO: How do we deal with side affects?
+    onSuccess();
```
peterp: The component transitions to another view once you've successfully logged in.

src/auth/reducer.js
```diff
@@ -0,0 +1,46 @@
...
+
+const auth = (
+  state = {
+    authToken: '',
+  },
+  action
+) => {
+  switch (action.type) {
+
+    case AUTH_LOGIN_REQUEST:
```
peterp: I remember that you split your data and UI reducers. I liked that, and would be interested in exploring it.

.eslintrc.js
```diff
@@ -2,5 +2,10 @@ module.exports = {
...
-    ]
-};
\ No newline at end of file
+    ],
+    "parserOptions":{
+      "ecmaFeatures": {
+        "experimentalObjectRestSpread": true
+      }
+    }
+};
```
justinvdm: Should we maybe add `babel-eslint` as a dev dep and use `parser: babel-eslint` here instead?

src/auth/ActivationContainer.js
```diff
@@ -0,0 +1,29 @@
...
+
+import { login } from './actions';
+import Login from './Login';
+
+const mapStateToProps = function mapStateToProps(state) {
+  return {
+    headerTitle: 'Activate Account',
+    buttonLabel: 'Activate',
```
justinvdm: Probably not relevant to this PR, but thought it would be worth asking about anyways. These seem like copy things, is there any way we keep copy things like this at their corresponding places in the markup?

Maybe where we use it, instead of`<Header title={this.props.headerTitle}>`, we could do `<Header title="Activate Account">`? Or am I missing a reason we need to do it this way?

.eslintrc.js
peterp: @justinvdm there's a note on the `babel-eslint` repo:

```
Note: You don't need to use babel-eslint if you are only using ES6/ES2015. ESLint actually supports ES6/ES7, JSX, and object rest/spread by default now.

At the moment, you'll need it if you use stuff like class properties, decorators, async/await, types.
```

Types? 8)

src/auth/actions.js
peterp: As you mentioned earlier we may be able to get react-router into redux; and therefore we could just dispatch an action that updates our state. Wooh!

.eslintrc.js
justinvdm: Oh, cool :) Didn't know it supports es7 by default now. I wonder why we have to explicitly enable object spread then though. Is our eslint up to date?

justinvdm:
:+1: looks good!

PR #23, opened 2016-08-08
Removed react-router-native

peterp:


justinvdm:
:+1: if we've grepped for stray `react-router-native` references and have noted down the routes we've removed somewhere (so we can bring them back later).

peterp:
@justinvdm I've made note of them in #24

PR #25, opened 2016-08-08
Upgraded React (~15.2.1) and React Native (^0.31.0)

peterp:


peterp:
@justinvdm Ideally I would want to keep the `android` folder out of the repo because it feels like a build step; but there's no way to just generate this folder with the `react-native`

justinvdm:
:+1:, I don't really understand the code in this PR, I'm assuming it is mostly generated stuff?

PR #26, opened 2016-08-10
Base Component Cleanup

peterp:


src/app/styles.js
```diff
@@ -0,0 +1,14 @@
...
+  brandonText: 'Brandon Text',
+  brandonTextMedium: 'Brandon Text_medium',
+  brandonTextBold: 'Brandon Text_bold',
+};
+
+export const fontWeight = {
+  normal: 'normal',
+  medium: '400',
+  bold: 'bold',
+};
```
justinvdm: My gut feel is that we should define base colours given in the pattern library here (just the colours themselves, not extending them to things like`borderColourGreen`), and then use those in our components' stylesheets. If you feel literals in all the stylesheets would be better, we can try that though.

src/profile/ProfilePicture.js
```diff
@@ -0,0 +1,44 @@
...
+  imageLabel: {
+    color: '#9fb1b3',
+    fontSize: 18,
+  },
+});
+
+// TODO: Picture path should be dynamic.
+const ProfilePicture = ({
+}) => {
+  const IMG_PROFILE_ADD = require('app/assets/Profile_Add.png');
```
justinvdm: I'm guessing there's a reason we use `require` here, but can't figure out why. Could you explain?

peterp: This component will also display the selected profile picture of the mentor. I will add some logic in that switches between including the selected profile picture or the placeholder.

Oooh! I just realised that react native added `defaultSource` as a property to `Image` components.

peterp:
@justinvdm This is ready for review; I'll split the others out into separate pull-requests.

src/profile/ProfilePicture.js
```diff
@@ -0,0 +1,43 @@
...
+  picturePath,
+}) => (
+  <BaseView>
+    <Toolbar title="Add a profile picture" />
+    <View style={styles.imageContainer}>
+      <Image
+        style={styles.image}
+        source={
+          picturePath && { uri: picturePath }
+          || require('app/assets/Profile_Add.png')
```
justinvdm: Could we make this an `import` at the top of the file?

justinvdm: Second thought: maybe our assets should also be constants we import from a constants module? Happy for us to do that in a separate PR.

src/components/TextInput.js
```diff
@@ -1,33 +1,72 @@
...
     return (
-      <View style={{
-        borderColor: '#dfe5e6',
-        borderBottomWidth: 1,
-        marginBottom: 24,
-      }}
+      <View
+        style={[
+          styles.container,
+          this.state.selected && styles.containerSelected,
```
justinvdm: What does react do if this evaluates to `false` if `this.state.selected` is false. I'm guessing it'll just ignore it?

peterp: Yup, it just ignores it.

src/profile/ProfilePicture.js
peterp: I'm a bit worried about that since it could mean that we're importing all our assets at launch, and not gaining anything by doing so.

I don't know exactly how react-native handles the the PNG: Is it a reference that's only loaded into memory once it's displayed, or is it in memory the moment it's imported?

justinvdm: Ah, thanks, good call. Should we just move this to the top as an `import` then?

peterp: 👍

justinvdm:
Looks good, liking the new approach to styles :) :+1: once the `requires` for `png`s are `import`s at the top of the file.

PR #28, opened 2016-08-10
Fix module casing

justinvdm:
We should be using `PascalCase` for modules exporting react components and lower case for everything else.


justinvdm:
@peterp ready for review.

peterp:
👍

PR #27, opened 2016-08-11
Refactor out api calls into programmatic api

justinvdm:
At the moment we make api calls directly in action creators. The plan is to refactor these calls into a programmatic api, then use them in the action creators.


justinvdm:
@peterp ready for review

src/auth/__tests__/actions-tests.js
```diff
@@ -1,51 +1,66 @@
...
+
+      const [_busy, action] = await capture(login('a@b.org', '1337'));
+
+      expect(action).toEqual({
+        type: AUTH_LOGIN_SUCCESS,
+        payload: { entities: { results: [23] } },
       });
+    });
+
+    it('should dispatch failure for non-empty results', async () => {
```
peterp: should dispatch failure for ~non-~empty results

src/auth/__tests__/reducers-test.js
```diff
@@ -0,0 +1,127 @@
...
+          },
+        },
+      });
+
+      expect(isLoading).toBe(true);
+    });
+
+    it('should clear the current error message', () => {
+      const { errorMessage } = reduce({
+        errorMessage: 'o_O',
```
peterp: 👍

src/auth/reducer.js
```diff
@@ -1,47 +1,51 @@
...
 
-const auth = (
-  state = {
-    authToken: '',
-  },
-  action
-) => {
+const auth = (state, action) => {
   switch (action.type) {
+    case AUTH_LOGIN_BUSY: {
```
peterp: I'm worried that merging `AUTH_LOGIN_BUSY` and `AUTH_SET_TOKEN` into the same action type could mean that you loose some of the ability to understanding what the type is doing.

Additionally, it may be necessary to remove the auth token when a user logs off, or when authentication failed (during any subsequent requests.)

src/auth/reducer.js
```diff
@@ -1,47 +1,51 @@
...
 
-    case AUTH_LOGIN_REQUEST:
       return {
         ...state,
+
         isLoading: true,
         errorMessage: '',
+
+        // TODO don't add `authToken` once we are using api in all actions
+        auth: conj(auth, { authToken: serializeAuth(email, password) }),
```
peterp: Could this be written as: 
`{...auth, { authToken: serializeAuth(email, password) } }`

src/config/index.js
```diff
@@ -0,0 +1,11 @@
+import { conj } from 'src/helpers';
+import base from 'src/config/base';
+import prd from 'src/config/prd';
+import dev from 'src/config/dev';
+import test from 'src/config/test';
+
+export default conj(base, {
```
peterp: Is this the same as writing:
`{...base, ...prd, ...dev, ...test}`

src/auth/reducer.js
justinvdm: > I'm worried that by merging AUTH_LOGIN_BUSY and AUTH_SET_TOKEN into the same action type it could mean that you loose some of the ability to understand what the type is doing just by glancing at the name.

The main reason for this is that the token setting is something temporary that will be moving away once the rest of the codebase is using the api. Maybe more importantly though, I think it is more useful to limit actions to describing application-specific events happening. `AUTH_SET_TOKEN` feels more like a step that a reducer does when an action happens, rather than an application event.

> The REQUEST/ SUCCESS / FAILURE nomenclature are already ingrained into the Redux community. I'm worried that a person may just assume that BUSY is just an indication of a request starting, and not much more.

Ah, I was thinking `REQUEST` was a remnant of the old way we were doing things where a middleware was making api requests. +1 to going with the convention, will change.

> Additionally, it may be necessary to remove the auth token when a user logs off, or when authentication failed in subsequent requests.

How about we only set auth on `AUTH_SUCCESS`, and add a `TODO` for handling an `AUTH_LOGOFF` action?

src/auth/reducer.js
justinvdm: Yep, I think the spread operator makes more sense here, will change.

Its the same thing really, I generally use `conj` when doing the same thing with the spread operator starts to look awkward. For example, when the things you're wanting to conjoin are already all variables, and you end up with lots of `...`s.

I'm not too phased about this though, happy to just use `...`s everywhere if you'd prefer that?

src/config/index.js
justinvdm: Its the same as writing:

```js
{
  ...base,
  {
    prd,
    dev,
    test,
  }[process.env.NODE_ENV || 'prd']
}
```

I think using `conj` here makes things a little more clearer, though I wouldn't mind using that ^ if you prefer.

src/auth/reducer.js
peterp: > The token setting is something temporary that will be moving away once the rest of the codebase is using the api. 

I think this is where my disconnect is happening as I don't yet know how that will happen, or what it'll look like. I'm happy to follow your lead though.

> How about we only set auth on AUTH_SUCCESS, and add a TODO for handling an AUTH_LOGOFF action?

Just to verify that I'm understanding this correctly:

`login(username, password)` would persist the username/ password with `AUTH_SUCCESS,` so that subsequent requests would use those credentials.

`AUTH_LOGOFF` would be dispatched if a request that requires authentication fails.

peterp: Also I don't have a particular preference for `BUSY` over `REQUEST`, but it may be worthwhile to follow the crowd.

Also `REQUEST` / `SUCCESS` / `FAILURE` are the same length! `ocd`

src/auth/reducer.js
```diff
@@ -1,47 +1,51 @@
...
       };
+    }
+
     case AUTH_LOGIN_FAILURE:
       return {
         ...state,
+
+        // TODO use constant for status type instead and put copy in component
```
peterp: 👍

src/auth/reducer.js
justinvdm: > login(username, password) would persist the username/ password with AUTH_SUCCESS, so that subsequent requests would use those credentials.

Yep

> AUTH_LOGOFF would be dispatched if a request that requires authentication fails.

I was thinking `AUTH_LOGOFF` (or maybe `AUTH_LOGOUT` rather) would be dispatched if the user to chooses to log out. Since we will only be adding the credentials to the state on successful login, we probably don't have to worry about resetting it on `AUTH_FAILURE`.

src/config/index.js
justinvdm: Erm, fixed syntax error in example above ^

src/auth/reducer.js
peterp: what if the username:password of the user is out of sync with the server and they haven't explicitly logged out?

I would like to lock the user out of the app, and show a dialog requesting that the user re-authenticates themselves.

justinvdm: It definitely sounds like a case we should handle, though I'm not sure I'm understanding the relevance of that case to the changes in this PR. What am I missing?

peterp: Sorry for rabbit holing Justin. I'm bringing it up because `SET_AUTH_TOKEN` was removed which was the mechanism that I was planning on using to automatically log the user off when authentication has failed.

I just want to make sure that you've got it in your mind that this should be something that we should be able to do.

peterp:
👍

PR #30, opened 2016-08-12
Call Notification and Overlay Dialogs.

peterp:


src/components/Button.js
```diff
@@ -2,40 +2,69 @@ import React, { PropTypes } from 'react';
...
+    color: COLORS.BUTTON_ALT_LABEL,
+  },
+  labelDisabled: {
+    color: COLORS.BUTTON_DISABLED_LABEL,
+  },
 });
 
 const Button = ({
-  label,
+  theme = 'orange',
```
peterp: We have two alternate styles for the button, orange and white

screen shot 2016-08-12 at 09 25 37

src/call/CallNotification.js
```diff
@@ -0,0 +1,72 @@
+import React, { PropTypes } from 'react';
```
peterp: screen shot 2016-08-10 at 18 04 39

src/components/CompletedOverlay.js
```diff
@@ -0,0 +1,102 @@
+import React, { PropTypes } from 'react';
```
peterp: screen shot 2016-08-11 at 15 02 08

src/components/LoadingOverlay.js
```diff
@@ -0,0 +1,60 @@
+import React, { PropTypes } from 'react';
```
peterp: screen shot 2016-08-12 at 09 28 23

src/constants/colors.js
```diff
@@ -0,0 +1,25 @@
+export default {
```
peterp: Once I have all the semantic constants in place I'll extract the colours into constants.

src/components/Button.js
```diff
@@ -2,40 +2,69 @@ import React, { PropTypes } from 'react';
...
+      style={[
+        styles.button,
+        styles[`button_${theme}`],
+        disabled && styles.buttonDisabled,
+      ]}
+    >
+      <Text
+        style={[
+          styles.label,
+          styles[`label_${theme}`],
```
justinvdm: As you suggested discussed on slack, should we use a constant for configuring themes here instead?

src/components/SpinningImage.js
```diff
@@ -0,0 +1,48 @@
...
+      outputRange: ['0deg', '360deg'],
+    });
+    return (
+      <Animated.Image
+        source={IMAGE_LOADER}
+        style={{
+          width: 80,
+          height: 80,
+          transform: [{ rotate: spin }],
+        }}
```
justinvdm: Would it be possible to add this to the stylesheet at the top?

src/components/SpinningImage.js
```diff
@@ -0,0 +1,48 @@
...
+    this.spin();
+  }
+
+  spin() {
+    this.state.spinValue.setValue(0);
+    Animated.timing(
+      this.state.spinValue,
+      {
+        toValue: 1,
+        duration: 2000,
```
justinvdm: Would be cool if we could use constants here.

justinvdm:
Left some comments, looks good :) I think we should note down somewhere that these components still need tests (so we can do that in a separate PR and not hold up this one).

Really like the generic components and the use of them from non-generic components! :)

Are the overlays going to be used by something outside of this PR?

PR #33, opened 2016-08-15
Semantic CONSTANTS for all colours

peterp:

PR #34, opened 2016-08-15
Remove `const styles` in Components, placed in `./styles`

peterp:


peterp:
@justinvdm this is ready for review.

justinvdm:
@peterp :+1:, probably makes sense for #33 to land first I'm guessing

PR #35, opened 2016-08-16
Use axios for api wrapper

justinvdm:
Axios uses xhr for making requests in react-native environments, which makes it a good fit for our api, since we need xhr for image uploads.


justinvdm:
@peterp ready for review

peterp:
👍

PR #37, opened 2016-08-17
Add context redux middleware

justinvdm:
The plan is to add a middleware that will use `getState()` to retrieve context to provide to any action creators that return a function. This will allow us to pass through auth details from our state without needing to pass around auth details all over our components and action creators.


justinvdm:
@peterp ready for review

src/stores/helpers.js
```diff
@@ -0,0 +1 @@
+export const getContext = ({ auth }) => ({ auth });
```
peterp: Is the idea that this may be expanded to return more than just auth?

justinvdm: I guess the less stuff we have here the better, so I'm hoping just auth would be enough, though yes if there were other global things that we needed to use often in all our action creators without needing to pass them around everywhere in the application (maybe id's of some sort?), then we could expand this.

It does look a bit useless, but the idea is to only be defining how we obtain the context from state in a single place, rather than doing this for all our action creators. If the way our state looked changed, we would only need to change how to obtain context from this in a single place.

peterp:
👍

PR #40, opened 2016-08-18
Use jest's toMatchSnapshot()

justinvdm:
We'd like to use this feature for testing that our components render their given properties according to what we've visually confirmed (and continue to as the codebase changes over time).

The plan is to require PRs with components and snapshot tests to have corresponding screenshots of the cases we are testing, to ensure that we have in fact visually confirmed that these cases are working as expected.


justinvdm:
label

justinvdm:
@peterp ready for review

peterp:
👍  WOOH!

PR #41, opened 2016-08-18
Remove unnecessary initialState() function

justinvdm:
I was initially thinking that `initialState()` would be a sensible place to define the initial values, but the defining the initial state in each reducer makes more sense.


justinvdm:
@peterp ready for review

peterp:
👍

PR #43, opened 2016-08-18
Change entity reducer to use a defined set of 'entity actions'

justinvdm:
At the moment we do this a bit implicitly by inspecting for an `entity` property in the action's payload.

Something like this seems a bit more explicit:

```js
// src/constants.js
const { PROFILE_FETCH_SUCCESS } from 'src/profile/constants';

const ENTITY_ACTIONS = [
  PROFILE_FETCH_SUCCESS,
];

// src/entitites/reducer.js
const reduceEntities = (state, action)) => includes(ENTITY_ACTIONS, action.type)
  ? merge(state, action.payload.entities)
  : state;
```

### DIFF
--- a/src/constants/actions.js
+++ b/src/constants/actions.js
@@ -8,4 +8,4 @@ const ACTIONS_WITH_ENTITIES = [
 
 export {
   ACTIONS_WITH_ENTITIES,
-}
+};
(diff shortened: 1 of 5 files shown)


justinvdm:
@peterp ready for review

peterp:
👍

PR #44, opened 2016-08-18
Clean up actions for schedule screen

justinvdm:
The plan is to add endpoints to our wrapper api for the schedule screen, and use this to clean up its actions and reducer.


justinvdm:
Opened PR against wrong ticket, making a new ticket, sorry for the noise.

PR #45, opened 2016-08-18
Less boilerplate for CRUD action creators

justinvdm:
The previous `generateActionCreator` made it difficult to see what actions a creator generated from it dispatched, and probably wasn't flexible enough for handling the nuances of some api endpoints, though it had value in that it meant we have 0 boilerplate.

The current approach we are taking sort of on the other end, intent is clearer and it is flexible, but we end up with lots of duplicated code when it comes to action creators for typical api crud.

We need to find a better approach somewhere in the middle.


justinvdm:
@peterp ready for review

peterp:
Nice! This really does strike a balance between CRUD 'n readability! 👍

justinvdm:
@peterp changed the approach to the way we test actions, a bit unsure of it but I think in the long run it is a better approach. Maybe best to discuss on a call.

PR #47, opened 2016-08-18
Authentication Views Navigation

peterp:


peterp:
@justinvdm this is ready for review.

src/navigation/auth/Navigator/mapRouteToComponent.js
```diff
@@ -0,0 +1,24 @@
...
+    case constants.AUTH_ROUTE_LANDING:
+      return <LandingContainer />;
+    case constants.AUTH_ROUTE_LOGIN:
+      return <LoginContainer />;
+    case constants.AUTH_ROUTE_ACTIVATE:
+      return <ActivationContainer />;
+    case constants.AUTH_ROUTE_RESET_PASSWORD:
+      return <ResetPasswordContainer />;
+    default:
+      return <View><Text>Route Not Found</Text></View>;
```
justinvdm: Not sure if we need to worry about this case too much, but if users land up on this view, they probably won't know what a route is, and its most likely a problem on our side I think. Maybe we need something like a message letting them know there's an issue with our app, and allow them to go back to the landing view, or something?

First prize would be a way of us being notified of bugs in the app, and letting the users know that we've been notified. Probably not worth the trouble for this case, but maybe we should have it in mind for similar cases that might be more likely.

src/navigation/auth/Navigator/index.js
```diff
@@ -0,0 +1,18 @@
...
+}) => (
+  <NavigationCardStack
+    navigationState={navigationState}
+    renderScene={(sceneProps) => mapRouteToComponent(sceneProps.scene.route.key)}
+  />
+);
+RootNavigation.propTypes = {
+  navigationState: PropTypes.any.isRequired,
+};
+export default RootNavigation;
```
justinvdm: If I'm groking this stuff correctly, I'm wondering if we could make this reusable if it took in `mapRouteToComponent` as a property, and moved it to `src/navigation` for other parts of the UI to use. Or would this component not really have much value in other parts of the UI?

src/navigation/auth/actions.js
```diff
@@ -0,0 +1,17 @@
...
+export const popAuthRoute = () => ({
+  type: constants.AUTH_NAVIGATION_POP,
+});
+
+export const pushLoginRoute = () =>
+  pushAuthRoute(constants.AUTH_ROUTE_LOGIN);
+export const pushResetPasswordRoute = () =>
+  pushAuthRoute(constants.AUTH_ROUTE_RESET_PASSWORD);
+export const pushActivateAccountRoute = () =>
+  pushAuthRoute(constants.AUTH_ROUTE_ACTIVATE);
```
justinvdm: Not very important, but it'd maybe be a little more consistent with the other actions if we used the helpers in `src/actionHelpers`.

```js
export const pushAuthRoute = dataAction(constants.AUTH_NAVIGATION_PUSH);
export const popAuthRoute = staticAction(constants.AUTH_NAVIGATION_POP);

export const pushLoginRoute = () =>
  pushAuthRoute({key: constants.AUTH_ROUTE_LOGIN});
export const pushResetPasswordRoute = () =>
  pushAuthRoute({key: constants.AUTH_ROUTE_RESET_PASSWORD});
export const pushActivateAccountRoute = () =>
  pushAuthRoute({key: constants.AUTH_ROUTE_ACTIVATE});
```

That said, those helpers seem more useful for reducing boilerplate for api-crud-related actions, could be overkill to use it here, happy either way, just thought it was worth mentioning.

src/navigation/auth/__tests__/reducer-tests.js
```diff
@@ -0,0 +1,56 @@
...
+} from 'src/navigation/auth/actions';
+
+import * as constants from 'src/navigation/auth/constants';
+import authReducer from 'src/navigation/auth/reducer';
+
+
+describe('navigation/auth/reducer', () => {
+  it('should have landing as the first route', () => {
+    // eslint-disable-next-line no-undefined
+    expect(authReducer(undefined, { type: 'ping?' })).toEqual({
```
justinvdm: Could we use `void 0` here to no avoid needing the eslint rule disabling?

src/navigation/auth/Navigator/mapRouteToComponent.js
peterp: That's a great idea. I'm going to create a base component for that.

src/navigation/auth/Navigator/index.js
peterp: It may be possible; there's some new stuff landing in RN 0.32.0 that allows you to define a custom header, which means that we can rip out the toolbar from the components, and have them rendered by the navigator.

I would like to see that in action, and build another navigator before I make the decision.

src/navigation/auth/__tests__/reducer-tests.js
```diff
@@ -0,0 +1,56 @@
...
+    };
+    expect(authReducer(state, popAuthRoute()))
+    .toEqual({
+      index: 0,
+      routes: [
+        { key: 'Ping?' },
+      ],
+    });
+  });
+});
```
justinvdm: Really cool that we can control our navigation state like this :)

src/navigation/auth/Navigator/index.js
justinvdm: Sounds good.

src/navigation/auth/actions.js
justinvdm: On second thought, maybe we need to add action helpers to `src/actionHelper` for routes that would allow us to do something like:

```js
export const popAuthRoute = popRoute(constants.AUTH_NAVIGATION_POP)

export const pushLoginRoute = pushRoute({
  type: constants.AUTH_NAVIGATION_PUSH,
  key: constants.AUTH_ROUTE_LOGIN
});
// ...
```

Seems like it might be useful if we are going to be doing similar things in other parts of the UI for routing. Not very important though, happy for us to leave it this way if you prefer it.

justinvdm:
Left minor comments, looks good :)

peterp:
@justinvdm ready for review

src/actionHelpers.js
```diff
@@ -24,3 +24,14 @@ export const dataAction = type => data => ({
   type,
   payload: data,
 });
+
+
+export const popRoute = type => () => ({
+  type,
+});
```
justinvdm: Sorry for only realizing this after suggesting this, but I think we can just use `staticAction` where we are currently using this instead.

justinvdm:
:+1: the last minor comment about `popAction -> staticAction` is resolved and travis is happy.

PR #49, opened 2016-08-19
Remove error messages from state for auth actions and reducer

justinvdm:
It would be a bit better if we changing a status type in our state, and switching on for displaying error messages.


justinvdm:
@peterp ready for review

src/auth/__tests__/Login-test.js
```diff
@@ -0,0 +1,17 @@
...
+
+
+describe('auth/Login', () => {
+  it('should draw not found statuses', () => {
+    expect(
+      <Login
+        status={authStatusNotFound()}
+        onLoginPress={noop}
+      />
+    ).toMatchSnapshot();
```
justinvdm: screen shot 2016-08-19 at 1 47 01 pm

peterp:
👍 🍰

PR #51, opened 2016-08-19
Remove no longer used code

justinvdm:
There are a few things we don't need anymore, including our api middleware and action generators.


justinvdm:
@peterp ready for review

peterp:
👍

PR #53, opened 2016-08-19
Add Activity API methods to client

justinvdm:


justinvdm:
@peterp ready for review

peterp:
I don't think there's any reason to update the activities (based of what I understand), but perhaps you saw something in the API?

Otherwise 👍

justinvdm:
> I don't think there's any reason to update the activities (based of what I understand), but perhaps you saw something in the API?

Ah, ya when I tested out the api, I noticed an `is_complete` field, which I thought we'd be setting manually at some point. Looking at the source, it looks like this is figured out by checking whether call feedback exists for the activity. Thanks, will remove.

PR #55, opened 2016-08-22
Add category list component

justinvdm:


justinvdm:
@peterp ready for review

src/categories/CategoryList/index.js
```diff
@@ -0,0 +1,42 @@
...
+  onCategoryPress,
+}) => (
+  <BaseView>
+      <View style={styles.header}>
+        <Text style={styles.text}>Relationship Stages</Text>
+      </View>
+
+      <View style={styles.list}>
+        {categories.map(({ id, title, color }) => (
+        <View
```
peterp: It's probably a good idea to use a `TouchableNativeFeedback` component here instead of the view; and add the `onPress` action to the parent.

src/categories/CategoryList/index.js
```diff
@@ -0,0 +1,42 @@
...
+import { BaseView, Link, Text } from 'src/components';
+import styles from './styles';
+
+
+const CategoryList = ({
+  categories,
+  onCategoryPress,
+}) => (
+  <BaseView>
+      <View style={styles.header}>
```
peterp: I think this is actually also a category, but it's poorly styled in the designs and looks like a header.

peterp: Actually, I just gave it another look and I'm wrong. It's a header. 🌴

src/categories/CategoryList/index.js
```diff
@@ -0,0 +1,42 @@
...
+import { View } from 'react-native';
+import { BaseView, Link, Text } from 'src/components';
+import styles from './styles';
+
+
+const CategoryList = ({
+  categories,
+  onCategoryPress,
+}) => (
+  <BaseView>
```
peterp: The baseView has a property called `statusBarBackgroundColor` which you can use to set the status bar to `#003035`

src/categories/CategoryList/index.js
```diff
@@ -0,0 +1,42 @@
...
+      <View style={styles.header}>
+        <Text style={styles.text}>Relationship Stages</Text>
+      </View>
+
+      <View style={styles.list}>
+        {categories.map(({ id, title, color }) => (
+        <View
+          key={id}
+          categoryId={id}
+          style={[styles.category, { backgroundColor: color }]}
```
justinvdm: The api seems to return colour information for categories, which I'm using here. I'm a bit worried about us not having any defaults for these though, seems quite possible for colours information to be left out when people add the categories in wagtail. If we do add a default scale, I think that should be outside of this PR, but thought it might be worth mentioning.

src/categories/CategoryList/styles.js
```diff
@@ -0,0 +1,24 @@
+import { StyleSheet } from 'react-native';
+import { FONT } from 'src/constants/styles';
+import colors from 'src/constants/colors';
+
+
+export default StyleSheet.create({
+  header: {
+    height: 96,
```
peterp: The height of the status bar is already `24`, so this height can be adjusted to `72`

peterp:
Besides those few little things; 👍

src/categories/CategoryList/index.js
justinvdm: Thanks. It looks like most of the headers in the app in the designs are of the height of this header and many are of this same colour. Thoughts on us making a header component with a theme for this colour?

peterp: There is a `Toolbar` component that should be able to deal with this; it
currently has a `backButton` that's always visible -- but we can make it so
that it isn't

Adding a theme option is a great idea!

justinvdm: Hmm, it seems like `Toolbar` expects a title prop rather than `children`. Quite a few of the headers in the designs don't really fit this. I think we need a more generic `Header` component, and either change `Toolbar` to use it, or have components choose to use one or the other.

peterp: +1 on using children instead. Also, feel free to rename to Header instead
of Toolbar if you feel the name is more appropriate.

src/components/Header/__tests__/Header-test.js
```diff
@@ -0,0 +1,30 @@
...
+import { Header, Title } from 'src/components';
+
+
+describe('Header', () => {
+  it('should render', () => {
+    expect(render(
+      <Header>
+        <Title>Microscopic Hummingbirds</Title>
+      </Header>
+    )).toMatchSnapshot();
```
justinvdm: screen shot 2016-08-23 at 7 44 37 pm

src/components/Header/__tests__/Header-test.js
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  it('should support style overriding', () => {
+    expect(render(
+      <Header style={{ backgroundColor: '#cc3333' }}>
+        <Title>Microscopic Hummingbirds</Title>
+      </Header>
+    )).toMatchSnapshot();
```
justinvdm: screen shot 2016-08-23 at 7 49 01 pm

src/components/Header/__tests__/Header-test.js
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  it('should support theming', () => {
+    expect(render(
+      <Header theme={{ header: { backgroundColor: '#33cccc' } }}>
+        <Title>Microscopic Hummingbirds</Title>
+      </Header>
+    )).toMatchSnapshot();
```
justinvdm: screen shot 2016-08-23 at 7 52 09 pm

src/categories/CategoryList/__tests__/CategoryList-test.js
```diff
@@ -0,0 +1,43 @@
...
+          title: 'Level',
+          color: '#97c13c',
+        }]}
+        {...props}
+      />
+    );
+  }
+
+  it('should render', () => {
+    expect(render(createComponent())).toMatchSnapshot();
```
justinvdm: screen shot 2016-08-23 at 8 10 16 pm

src/components/Title/__tests__/Title-test.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { Title } from 'src/components';
+
+
+describe('Title', () => {
+  it('should render', () => {
+    expect(render(
+      <Title>Microscopic Hummingbirds</Title>
+    )).toMatchSnapshot();
```
justinvdm: screen shot 2016-08-23 at 7 44 37 pm

src/components/Title/__tests__/Title-test.js
```diff
@@ -0,0 +1,23 @@
...
+  });
+
+  it('should support style overriding', () => {
+    expect(render(
+      <Title style={{ fontSize: 22 }}>Microscopic Hummingbirds</Title>
+    )).toMatchSnapshot();
```
justinvdm: screen shot 2016-08-23 at 7 48 32 pm

src/components/Title/__tests__/Title-test.js
```diff
@@ -0,0 +1,23 @@
...
+  });
+
+  it('should support theming', () => {
+    expect(render(
+      <Title theme={{ title: { color: '#eee' } }}>Microscopic Hummingbirds</Title>
```
justinvdm: screen shot 2016-08-23 at 7 54 09 pm

justinvdm:
@peterp ready for re-review

justinvdm:
Snapshots after [this commit](https://github.com/praekeltfoundation/mentorship-droid/pull/55/commits/6a50059e520cf840753404994d8f5c34e90efcd3):
screen shot 2016-08-23 at 9 27 24 pm
screen shot 2016-08-23 at 9 39 23 pm
screen shot 2016-08-23 at 9 43 51 pm

src/categories/CategoryList/index.js
```diff
@@ -0,0 +1,39 @@
...
+import { BaseView, Header, Title } from 'src/components';
+import styles from './styles';
+
+
+const CategoryList = ({
+  categories,
+  onCategoryPress,
+}) => (
+  <BaseView>
+      <Header theme={Header.themes.dark}>
```
peterp: I like the way you did this; going to incorporate the same thing for the button.

src/categories/CategoryList/index.js
```diff
@@ -0,0 +1,39 @@
...
+        <Title theme={Title.themes.light}>Relationship Stages</Title>
+      </Header>
+
+      <View style={styles.list}>
+        {categories.map(({ id, title, color }) => (
+        <TouchableNativeFeedback
+          key={id}
+          categoryId={id}
```
peterp: Is this `categoryId` used for something?

src/categories/CategoryList/index.js
justinvdm: \o/

src/categories/CategoryList/index.js
justinvdm: Ya, for finding it in tests (we can't use `key` to do that). We could probably get away with finding on `TouchableNativeFeedback` or something else, but that makes the tests a bit brittle for when the view changes.

src/categories/CategoryList/index.js
```diff
@@ -0,0 +1,39 @@
...
+          onPress={() => onCategoryPress(id)}
+        >
+          <View style={[styles.category, { backgroundColor: color }]}>
+            <Title theme={Title.themes.light}>{title}</Title>
```
peterp: I'm a bit worried about using `Title` here; as it's not exactly clear, from the components name, what benefit is has over `Text` other than some arbitrary default styling.

My fear is that we may start using `Title` because it's a medium weighted font; and then we may overwrite the size, and several other properties to get it to conform. It could make things messy and unpredictable.

I would prefer it if `Text` is used, or if `Title` couldn't accept any additional `style` attributes.

justinvdm: I was a bit worried about this. I think it has semantic value -- these are category titles. I also think it is why they are styled liked titles in the designs. Will change to `Text` though.

src/components/Header/index.js
```diff
@@ -0,0 +1,27 @@
...
+  theme = {},
+}) => (
+  <View style={[styles.header, theme.header, style]}>
+    {children}
+  </View>
+);
+
+
+Header.propTypes = {
+  style: PropTypes.object,
```
peterp: I've done a bit of research into how these fairly common attributes should be validated:

```
// Use this for style
  style: PropTypes.any,

  // Use this for children
  children: PropTypes.node,

  // Use this for nested children
  children: PropTypes.oneOfType([
    PropTypes.arrayOf(PropTypes.node),
    PropTypes.node,
  ]),
```
 http://blog.tylerbuchea.com/react-proptypes/

What do you think?

justinvdm: +1, thanks, searched for something like this, to no avail.

justinvdm:
Ready for another review.

PR #58, opened 2016-08-23
Fix snapshot testing

justinvdm:
We weren't snapshot testing properly it seems :/


justinvdm:
@peterp ready for review

package.json
```diff
@@ -12,41 +12,48 @@
...
-    "babel-preset-react-native": "^1.9.0",
     "enzyme": "^2.4.1",
     "eslint": "^2.3.0",
     "eslint-config-airbnb": "^6.1.0",
     "eslint-plugin-react": "^4.2.1",
     "isomorphic-fetch": "^2.2.1",
-    "jest-cli": "^13.0.0",
-    "react-addons-test-utils": "^15.3.0",
-    "react-dom": "^15.3.0",
-    "react-native-mock": "^0.2.6",
```
justinvdm: Apparently we don't need react-native-mock for our enzyme tests, though I'm not actually sure why. Maybe something that is used as part of snapshot testing (jest-react-native?) is doing the necessary mocking there?

PR #61, opened 2016-08-24
Fix env configs

justinvdm:
We need to use `__DEV__` since react native sets `NODE_ENV` itself depending on its value.


justinvdm:
@peterp ready for review

peterp:
👍

PR #56, opened 2016-08-24
Add base component for category detail screen

justinvdm:
The category detail screen is a tabbed view. This ticket is for creating the base component with its header and tabs, and stubs for its tabs' views.


justinvdm:
@peterp ready for review

src/categories/Category/__tests__/Category-test.js
```diff
@@ -0,0 +1,53 @@
...
+  });
+
+  it('should render the currently active tab', () => {
+    let el;
+    el = render(createComponent({ initialPropType: TAB_ABOUT }));
+    expect(el).toMatchSnapshot();
+
+    el = render(createComponent({ initialPropType: TAB_ACTIVITIES }));
+    expect(el).toMatchSnapshot();
+  });
```
justinvdm: screen shot 2016-08-24 at 5 28 33 pm
screen shot 2016-08-24 at 5 28 39 pm

src/categories/Category/index.js
```diff
@@ -0,0 +1,127 @@
...
+
+Category.propTypes = {
+  category: PropTypes.object.isRequired,
+  initialActiveTab: tabPropType,
+};
+
+
+CategoryTab.propTypes = {
+  tabId: PropTypes.string.isRequired,
+  isActive: PropTypes.bool.isRequired,
```
peterp: There's a convention where you don't have to specify a prop that's false. 

I know we're computing this, but I wondering if it isn't worth it to stick to the convention by not making it required; `active` may also be more suitable name than `isActive`

peterp:
👍 , wow, it's not much code!

src/categories/Category/index.js
justinvdm: I'm not sure I agree with that convention. In terms of what the component expects, this is a necessary parameter, rather than an optional one that should default to a falsy value. That approach could lead to unintended behaviour: For example, if the component using it was changed by and by accident no longer specified this property. I'd kind of prefer we left it compulsory.

Not to be belligerent, but I kind of also disagree on `active` being more suitable than `isActive`, sorry :) Sometimes its difficult to tell whether something is a flag or a method. For example, by this convention, `open` could be a method or a flag representing that something is open. We don't have that problem for the case of `active`, but its nice to be consistent with the naming we use for flags.

peterp: 👍 , on keeping it required.

Non-belligerance noted, `/me adds +2 to healthy discussion`

I always tend to read `isActive` / `isDisabled` as a getter rather than a setter, and when writing JSX it seems nicer to write:

`<Button disabled title="Bad button" />`
vs.
`<Button isDisabled title="Bad button" />`

justinvdm: Hmm, that kind of made me realise that the builtin react native components probably follow that convention already, so the convention we should be using is sort of chosen for us already if we want to be consistent with the way they're doing things.

Not really a fan, but seems worth sticking with that convention, will change to `active`.

justinvdm:
@peterp ready for re-review.

peterp:
👍

PR #65, opened 2016-08-25
Root Navigation

peterp:


src/config/base.js
```diff
@@ -1,3 +1,3 @@
 export default {
-  API_URL: '',
+  API_URL: 'http://10.0.3.2:8000/mentor',
 };
```
justinvdm: The base config is where defaults common to all configs go. I think this stuff should move to `development.js`.

src/config/production.js
```diff
@@ -0,0 +1,3 @@
+export default {
+  API_URL: 'http://qa-gem-mentorship.za.prk-host.net/mentor',
+};
```
justinvdm: I think we should remove this config from our repo, and not keep it or `development.js` in the repo. I thought I'd added them to `.gitignore`, but it seems I didn't.

src/navigation/routes/index.js
```diff
@@ -0,0 +1,40 @@
...
+      return <WelcomeContainer />;
+    case constants.ROUTE_ONBOARDING_PROFILE_PICTURE:
+      return <ProfilePictureContainer />;
+    case constants.ROUTE_ONBOARDING_CAMERA_ROLL:
+      return <CameraRollContainer />;
+
+
+    default:
+      return <NotFound />;
+  }
```
justinvdm: I'm wondering if we should rather declare these mappings statically. For e.g,

```js
const COMPONENT_ROUTES = {
  ROUTE_AUTH_ACTIVATION: ActivationContainer
  // ...
};

const Router = key => {
  const Container = COMPONENT_ROUTES[key] || NotFound;
  return <Container />;
};


Router.COMPONENT_ROUTES = COMPONENT_ROUTES;


export default Router;
```

Seems like less can go wrong that way, considering these mappings are likely to grow.

Happy either way though.

peterp: 👍  I'll do that quickly.

src/onboarding/actions.js
```diff
@@ -0,0 +1,15 @@
...
+  payload: {
+    profilePicture: path,
+  },
+});
+
+export const changeProfileImageAndPreview = path => dispatch => Promise.resolve()
+  .then(() => changeProfileImage(path))
+  .then(dispatch)
+  .then(popRoute)
+  .then(dispatch);
```
justinvdm: I'm thinking more and more that adding action creators for our route changes wasn't a good idea. I feel like our action should be describing things that happened in our application, and our reducers should be changing state to control what we do about those actions. In that sense, `pushRoute` and `popRoute` feel more like they're the reducers job.

If we wanted to take that approach, this action creator would only be changing the profile image, and we would pop the route in the reducer when handling `ONBOARDING_CHANGE_PROFILE`.

I'd really like us to take an approach like that instead, but if that is going to be huge change, then maybe it is more sensible keeping the current approach. What are your thoughts?

src/profile/reducer.js
```diff
@@ -1,6 +1,12 @@
 import * as constants from 'src/profile/constants';
 import * as statuses from 'src/profile/statuses';
 
+import { getProfileId } from 'src/auth/reducer';
+
+
+export const getProfile = (state, profileId) => state.entities.profiles[profileId];
+export const getProfileName = (state) => getProfile(state, getProfileId(state)).first_name;
```
justinvdm: I'm wondering if we continued with this approach, whether we'd end up with `getProfileFoo` for each `foo` in a profile model (and similarly for other models). Would it maybe be less awkward if we just let the things needing this information use `getProfile`, and then pick things off the model directly.

(If we're worried about a changing api, we could have getProfile include a parsing step to safeguarding us against that.)

peterp: So the reason why I did this was because of `first_name`, `last_name`, etc... I didn't want that to bleed out into the open. Any ideas on how to make that cleaner?

src/onboarding/actions.js
peterp: I'm a little bit cautious on the approach above mainly because I don't know how to guarantee the order of actions. Do you think we could place the navigation reducer last?

src/profile/reducer.js
justinvdm: Hmm, I think `api/request()` should have a (switch-off-able) parse step to camelcaseify and a serialize step to snakecaseify. For now, could we add an extra `firstName` field to `getProfile` with a `TODO` for removing it, and I'll make a PR for that conversion. Thoughts?

peterp: 👍

src/onboarding/actions.js
justinvdm: If I'm understanding correctly, it seems [we don't have to worry about that case](http://stackoverflow.com/questions/35929543/how-to-define-the-order-of-reducers-execution-in-redux) :)

justinvdm:

peterp:
@justinvdm ready for rereview

src/profile/reducer.js
```diff
@@ -1,6 +1,18 @@
 import * as constants from 'src/profile/constants';
 import * as statuses from 'src/profile/statuses';
 
+export const getProfile = (state, profileId) => {
+  if (!profileId) return {};
```
justinvdm: If this case were to happen, it'd likely be a bug. I think it'd be better to just throw an error. At a later stage, we can try catch the error (see what I did there ;), and use it to display something to the user that there is an error on our side.

src/profile/reducer.js
```diff
@@ -1,6 +1,18 @@
...
+
+  const {
+    first_name: firstName,
+  } = state.entities.profiles[profileId] || {};
```
justinvdm: Wondering if it'd be better for us to land #68 first and then remove this. Seems like less work that way. Happy for this to land before it though, and then we make a new PR for removing this workaround, if you prefer that approach.

justinvdm:
Left two more minor comments, looks good :)

src/navigation/routes/actions.js
```diff
@@ -0,0 +1,12 @@
...
+import * as constants from 'src/navigation/constants';
+
+import { AUTH_LOGIN_SUCCESS } from 'src/auth/constants';
+import { ONBOARDING_CHANGE_PROFILE } from 'src/onboarding/constants';
+
+
+export default fromPairs([
+  [AUTH_LOGIN_SUCCESS, makePushRoute(constants.ROUTE_ONBOARDING_WELCOME)],
+  [ONBOARDING_CHANGE_PROFILE, popRoute],
+]);
```
justinvdm: Sugoi desu! 🍰 !

src/profile/reducer.js
peterp: hahaha, you're funny ;)

src/profile/reducer.js
peterp: Ok, let's land #68!

justinvdm:
:+1: when travis is happy

PR #68, opened 2016-08-25
Normalize case for api requests and responses

justinvdm:
So we don't have snakecase in our camelcase.


justinvdm:
@peterp ready for review

peterp:
👍

PR #66, opened 2016-08-25
Add section component

justinvdm:
These are used in multiple places in the app, including the category about page and call notes.

PR #70, opened 2016-08-26
Use styles prop for themes

justinvdm:
Right now we do `<Foo theme={Foo.themes.bar}>`. It would be cool if we could use `<Foo style={Foo.themes.bar}>`.


justinvdm:
@peterp depends on #66, otherwise ready for review.

I haven't changed `Button`, it is likely quite tied to other branches yet to land. We should update it once those have landed.

peterp:
👍

PR #71, opened 2016-08-26
Added `themes` and `layouts` to Button

peterp:
I've brought this inline with our current setup:

screen shot 2016-08-26 at 15 46 46
screen shot 2016-08-26 at 15 46 35
screen shot 2016-08-26 at 15 46 16


peterp:
@justinvdm this is ready for review

src/components/Button/index.js
```diff
@@ -41,12 +40,26 @@ const Button = ({
...
-  theme: PropTypes.oneOf([THEME_DEFAULT, THEME_WHITE]),
+  theme: PropTypes.oneOf([
+    themes.default,
+    themes.light,
+    themes.transparent,
+  ]),
+  layout: PropTypes.oneOf([
+    layouts.default,
+    layouts.inline,
+  ]),
```
justinvdm: Wondering if we do in fact want to limit layouts and themes to ones provided by the component. Seems possible that we'd want to provide a special-cased theme or layout. Thoughts?

justinvdm:
👍 !

src/components/Button/index.js
peterp: Yeah, agreed.

PR #73, opened 2016-08-26
Button should accept `children` prop

peterp:
So that images can be Rendered.


peterp:
@justinvdm ready for review

peterp:
@justinvdm fixed merge conflict, ready for rereview.

justinvdm:
:+1:

Considering we now need to uppercase explicitly for buttons, I'm wondering whether my suggestion to go this route was a good idea.

peterp:
@justinvdm Perhaps I should check the type, and uppercase it if it's a string.

PR #77, opened 2016-08-29
Change directory structure

justinvdm:
At the moment we use `src/<section-of-ui>/{reducer,actions,<containers>,<components>}`, but we've found the UI is more interlinked than this. For e.g. we may use the same view in different parts of the UI. We're changing to `src/{reducers,actions,containers,components,views}/`.


justinvdm:
@peterp ready for review

peterp:
👍

PR #78, opened 2016-08-30
Onboarding: Occupation

peterp:
@justinvdm I decided to keep the state in the reducer instead of the component; since the data is to be used at the end to perform the API call I felt that it's best to remove additional moving parts.

screen shot 2016-08-30 at 15 15 44

screen shot 2016-08-30 at 15 15 36

Ready for review.


peterp:
@justinvdm ready for review

src/views/OnboardingOccupation/index.js
```diff
@@ -0,0 +1,54 @@
...
+  jobTitle = '',
+  jobSector = '',
+  onChangeText,
+  onBackPress,
+  onNextPress,
+}) => {
+  const isPaginationDisabled = jobTitle.length === 0 || jobSector.length === 0;
+  return (
+    <BaseView>
+      <ProgressBar completed={0.3 / 0.7} />
```
justinvdm: Would this be clearer as `3 / 7` if there are `7` steps? Or is there a reason for the floats. Sorry, a bit bikesheddy.

src/routes.js
```diff
@@ -4,16 +4,27 @@ import * as constants from 'src/constants/navigation';
...
+
+
 import WelcomeContainer from 'src/containers/WelcomeContainer';
 import ProfilePictureContainer from 'src/containers/ProfilePictureContainer';
 import CameraRollContainer from 'src/containers/CameraRollContainer';
+import OnboardingOccupationContainer from 'src/containers/OnboardingOccupationContainer';
+import OnboardingMotivationContainer from 'src/containers/OnboardingMotivationContainer';
+// import OnboardingInspirationContainer from 'src/containers/OnboardingInspirationContainer';
+// import OnboardingThreeWordsContainer from 'src/containers/OnboardingThreeWordsContainer';
+// import OnboardingSkillsContainer from 'src/containers/OnboardingSkillsContainer';
```
justinvdm: Should we just remove these for now? Unless there's a reason we want to keep these commented out bits around.

src/containers/OnboardingOccupationContainer.js
```diff
@@ -0,0 +1,16 @@
...
+
+export default connect(
+  (state) => ({
+    jobTitle: state.onboarding.jobTitle,
+    jobSector: state.onboarding.jobSector,
+  }),
+  {
+    onChangeText: updateProfile,
+    onBackPress: popRoute,
+    onNextPress: pushOnboardingMotivation,
```
justinvdm: Not very important, but I feel like it would be more in line with the approach we discussed last week if these dispatched actions like 'ONBOARDING_OCCUPATION_BACK` and `ONBOARDING_OCCUPATION_NEXT`, and then we map those to new routes in `actions/routes`.

It would accomplish the same thing, but in the end, instead of needing to look inside each container to figure out how the screens flow from one to the other, we would end up with a set of mappings which would make it clearer how these screens flow and how application changes result in navigation changes, e.g:

```js
export default fromPairs([
  [AUTH_LOGIN_SUCCESS, makePushRoute(constants.ROUTE_ONBOARDING_WELCOME)],
  [ONBOARDING_CHANGE_PROFILE, popRoute],
  [ONBOARDING_OCCUPATION_BACK, makePushRoute(constants.ROUTE_ONBOARDING_MOTIVATION],
  // ...
]);
```

justinvdm:

justinvdm:
:+1: once the comments and `0.3 / 0.7` comments are resolved and travis is happy. We can discuss and possibly change the route stuff in a separate PR if you preferp.

src/containers/OnboardingOccupationContainer.js
justinvdm: Just realised we can have a `BACK` action and use that across the application, which would change `actions/routes` to look something like this if we take the approach I'm suggesting above:

```js
export default fromPairs([
  [AUTH_LOGIN_SUCCESS, makePushRoute(constants.ROUTE_ONBOARDING_WELCOME)],
  [BACK, popRoute],
  [ONBOARDING_OCCUPATION_NEXT, makePushRoute(constants.ROUTE_ONBOARDING_MOTIVATION)],
  // ...
]);
```

justinvdm:
:+1: wtih

PR #79, opened 2016-08-30
Add chain middleware

justinvdm:
We need a way to 'chain' actions onto other actions, so that once the source action has been dispatched, the target action gets dispatched. This will allow us to co-ordinate how route changes and api fetches happen as a result of other actions (for e.g. actions for fetching data and routing to the journey page when an `AUTH_LOGIN_SUCCESS` action is dispatched). This is basically what we are doing already for routing, but allows us to use thunks.

This approach seems better than the other alternatives I could think of:
  - just dispatching a fetch action when a container is updated: the view layer doesn't really have anything to do with this stuff, so ideally it shouldn't be involved
  - constructing awkward action creators: for e.g., instead of `loginAndFetchFromApiAndRouteToJourneyPage`, we can have a `login` action creator, and then, in a module dedicated to routing things, use a more declaritive approach to map the resulting `AUTH_LOGIN_SUCCESS` to action creators for feching from the api and for routing
  - using one of the existing approaches for this (redux-saga, redux-loop): these seem a bit flamboyant, and would require changing what we currently have quite a bit, probably more work than we have time for


justinvdm:
@peterp ready for review

justinvdm:
@peterp also, let me know if you have a better name for this than `chainMiddleware` :)

peterp:
👍 ! I think chain middleware is fine.

PR #80, opened 2016-08-30
Fix content middleware

justinvdm:
The context middleware currently expects to get `dispatch` and `getState` as positional arguments rather than named arguments. It also tries to do redux-thunk's job, rather than wrapping thunks and passing them to redux-thunk.


justinvdm:
@peterp ready for review

peterp:
👍

PR #81, opened 2016-08-30
Loading screen action and view

justinvdm:
This is for adding an action creator for performing the initial load, and a corresponding loading view (uses the loading overlay component).

Was initially planning on adding the routing changes and hooking up of the `load()` action creator in this PR, but I think I need a better idea of how that should fit in with the current onboarding routes, and those changes seemed separate enough to leave for a separate PR.


justinvdm:
@peterp ready for review

peterp:
Very, very cool. 👍

justinvdm:
@peterp made some changes (some of which we just discussed), ready for review.

peterp:
👍

PR #82, opened 2016-08-31
MultiLineTextInput Component

peterp:
screen shot 2016-08-31 at 10 34 21

@justinvdm Ready for Review

### DIFF
--- a/src/components/MultiLineTextInput/index.js
+++ b/src/components/MultiLineTextInput/index.js
@@ -27,7 +27,6 @@ export default class MultiLineTextInput extends React.Component {
         <TextInput
           {...this.props}
           multiline
-          enablesReturnKeyAutomatically
           onChange={(event) => this.setState({
             height: event.nativeEvent.contentSize.height,
           })}
--- a/src/components/MultiLineTextInput/index.js
+++ b/src/components/MultiLineTextInput/index.js
@@ -5,11 +5,14 @@ import styles from './styles';
 
 
 export default class MultiLineTextInput extends React.Component {
-  constructor() {
-    super();
+  constructor(props) {
+    super(props);
+
+    const { value = '' } = props;
+
     this.state = {
       height: 0,
-      charCount: 0,
+      charCount: value.length || 0,
     };
   }
 
@@ -50,6 +53,7 @@ export default class MultiLineTextInput extends React.Component {
 
 
 MultiLineTextInput.propTypes = {
+  value: PropTypes.string,
   onChangeText: PropTypes.func,
   maxLength: PropTypes.number,
 };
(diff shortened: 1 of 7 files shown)


src/components/MultiLineTextInput/index.js
```diff
@@ -1,29 +1,56 @@
...
-            height: event.nativeEvent.contentSize.height + 8,
-          })
-        }
-        style={{ height: Math.max(28, this.state.height) }}
-      />
+      <View>
+        <TextInput
+          {...this.props}
+          multiline
+          enablesReturnKeyAutomatically
```
justinvdm: This prop seems ios specific.

src/components/MultiLineTextInput/index.js
```diff
@@ -1,29 +1,56 @@
...
+          onChange={(event) => this.setState({
+            height: event.nativeEvent.contentSize.height,
+          })}
+          onChangeText={(text) => {
+            this.setState({
+              charCount: text.length,
+            });
+            if (this.props.onChangeText) {
+              this.props.onChangeText(text);
+            }
```
justinvdm: Can we add a test for this?

src/components/MultiLineTextInput/index.js
```diff
@@ -1,29 +1,56 @@
...
     };
   }
 
   render() {
+    let charCount;
+    if (this.props.maxLength) {
+      charCount = (
+        <Text style={styles.charCount}>
+          {this.state.charCount}/{this.props.maxLength}
+        </Text>);
```
justinvdm: I'm wondering what should happen if they go over the char count. Maybe we should give some feedback about that by changing the colour of the char count text to a warning colour?

Sort of related -- I guess the component using it also needs to know about this so that it can disable its submit/continue/next button or something for this case. We could maybe have `onOverLimit` and `onUnderLimit` callbacks, but that feels a bit weird. Maybe having the component using it check for this itself using `onTextChange` is better?

justinvdm:

src/components/MultiLineTextInput/index.js
peterp: 👍

src/components/MultiLineTextInput/index.js
peterp: `maxLength` is a property on `TextInput`, which restricts the number of characters. It's not possible to go over the limit.

https://facebook.github.io/react-native/docs/textinput.html#maxlength

justinvdm: Woops, missed that, thanks :)

justinvdm:
:+1: once the test and ios prop comments are resolved and travis is happy.

src/components/MultiLineTextInput/index.js
peterp: I "kinda" added a test for this with an initial value; but I don't know how to simulate text.

justinvdm:
:+1:!

PR #83, opened 2016-08-31
Onboarding: Motivation

peterp:
screen shot 2016-08-31 at 12 25 45

@justinvdm Ready for review.

### DIFF
--- a/src/views/OnboardingMotivation/index.js
+++ b/src/views/OnboardingMotivation/index.js
@@ -1,0 +1,49 @@
+import React, { PropTypes } from 'react';
+import { View } from 'react-native';
+import { BaseView, Header, Text, Pagination, ProgressBar,
+  MultiLineTextInput } from 'src/components';
+import styles from './styles';
+
+
+const OnboardingMotivation = ({
+  jobMotivation = '',
+  onChangeText,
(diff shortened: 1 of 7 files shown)


src/views/OnboardingMotivation/index.js
```diff
@@ -0,0 +1,49 @@
...
+import styles from './styles';
+
+
+const OnboardingMotivation = ({
+  jobMotivation = '',
+  onChangeText,
+  onBackPress,
+  onNextPress,
+}) => {
+  const isPaginationDisabled = jobMotivation.length <= 10;
```
justinvdm: This min length is maybe a bit unclear to users. If they entered a short motivation, they might be confused about why the next button is still disabled. I think we should either have the minimum be `1` char, or should make the minimum limit clear in some way.

justinvdm:
👍 once the min char comment is resolved. I think we still need a test for asserting that `onChangeText` is called when the multiline text input is changed, but that should probably be a todo since we can't test that properly at the moment.

src/views/OnboardingMotivation/index.js
peterp: Good point; I'm going to remove that check.

PR #84, opened 2016-08-31
Onboarding: Own Mentor

peterp:
@justinvdm Ready for review.

screen shot 2016-08-31 at 12 43 29


justinvdm:
:+1:

Noting that we need a test for the `onTextChange`, but that we don't have it here since we can't test `TextInput` changes at the moment.

PR #85, opened 2016-08-31
Onboarding: Three Words

peterp:
I'll probably need to build a more robust "tag" component, but I don't have the bandwidth to do that right now, and it feels like something that'll take a lot of time, require a ton of testing, and doesn't really have very good value. 

But it may be a nice open source component. :D

screen shot 2016-08-31 at 12 55 38

@justinvdm Ready for review.

### DIFF
--- a/src/views/OnboardingThreeWords/index.js
+++ b/src/views/OnboardingThreeWords/index.js
@@ -11,6 +11,7 @@ const OnboardingOwnMentor = ({
   onBackPress,
   onNextPress,
 }) => {
+  // TODO: Create tag text input component
   const isPaginationDisabled = threeWords.split(' ').length < 3;
   return (
     <BaseView>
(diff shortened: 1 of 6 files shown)


src/views/OnboardingThreeWords/index.js
```diff
@@ -0,0 +1,49 @@
...
+import styles from './styles';
+
+
+const OnboardingOwnMentor = ({
+  threeWords = '',
+  onChangeText,
+  onBackPress,
+  onNextPress,
+}) => {
+  const isPaginationDisabled = threeWords.split(' ').length < 3;
```
justinvdm: I think we need something more robust than this. For e.g. the user could separate with multiple spaces, or other  characters. If we want it to only be spaces, we should mention that in the copy. Maybe we need a helper function that splits on whitespace and commas (and mention in the copy that they can separate with spaces or commas)?

Happy with just adding a TODO here for something like that if a new PR for that would be better.

justinvdm:
Other than the one comment, :+1:wtih.

PR #88, opened 2016-08-31
Onboarding: Done/ Saving Screen

peterp:
@justinvdm ready for review


src/views/OnboardingSaving/index.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { OverlayLoading } from 'src/components';
+
+const OnboardingSaving = () => {
+  return (
+    <OverlayLoading title="Saving Profile..." />
```
justinvdm: Wondering if we should make it clear in the copy that this is the end at the moment (otherwise the people testing will wait forever and count it as a bug).

justinvdm:
:+1:

PR #89, opened 2016-08-31
Move navigation logic to reducer

justinvdm:
At the moment we do it via actions, which starts to get weird when conditional routing is necessary. We could push this logic to containers, but routing logic isn't really a container's job.


justinvdm:
@peterp ready for review.

src/reducers/navigation.js
```diff
@@ -1,29 +1,53 @@
...
+      return forward(state);
+
+    case landing.LANDING_GET_STARTED:
+      return push(state, createRoute(routes.ROUTE_AUTH_ACTIVATION));
+
+    case landing.LANDING_LOGIN:
+      return push(state, createRoute(routes.ROUTE_AUTH_LOGIN));
+
+    case auth.AUTH_LOGIN_SUCCESS:
+      return reset(state, [createRoute(routes.ROUTE_ONBOARDING_WELCOME)]);
```
peterp: I'm wondering what kind of transition animation `reset` does, might be worthwhile to check it out.

peterp: It transitions backwards, when it should transition forwards. It's a bit awkward.

src/reducers/navigation.js
```diff
@@ -1,29 +1,53 @@
...
+
+    case onboarding.ONBOARDING_START_PROFILE:
+      return reset(state, routes.ONBOARDING_STEPS.map(createRoute), 0);
```
peterp: Same transition problem here. It's a hard cut directly to the next screen, instead of an animation.

src/constants/onboarding.js
```diff
@@ -1,10 +1,16 @@
+const ONBOARDING_START_PROFILE = 'ONBOARDING_START_PROFILE';
+const ONBOARDING_CHOOSE_PROFILE_PICTURE = 'ONBOARDING_CHOOSE_PROFILE_PICTURE';
+
 const ONBOARDING_UPDATE_PROFILE_PICTURE = 'ONBOARDING_UPDATE_PROFILE_PICTURE';
 const ONBOARDING_UPDATE_PROFILE = 'ONBOARDING_UPDATE_PROFILE';
 
 const ONBOARDING_OCCUPATION_NEXT = 'ONBOARDING_OCCUPATION_NEXT';
```
peterp: Was this left here by accident?

src/reducers/navigation.js
```diff
@@ -1,29 +1,53 @@
...
+
+    case onboarding.ONBOARDING_CHOOSE_PROFILE_PICTURE:
+      return insert(state, createRoute(routes.ROUTE_ONBOARDING_CAMERA_ROLL));
```
peterp: This adds a route to the end of the stack, when it should`index + 1`, is there any way to specify where it should be added?

peterp: I just had a look at the function, it appears that it should be doing that, but I'm not seeing those results.

src/reducers/navigation.js
peterp: I'm wondering if there's a way to specify transitions.

src/reducers/navigation.js
```diff
@@ -1,29 +1,53 @@
...
 
-    case constants.NAVIGATION_POP:
-      return NavigationStateUtils.pop(state);
+    case onboarding.ONBOARDING_UPDATE_PROFILE_PICTURE:
+      return pop(state);
```
peterp: This may have to be `back` rather than `pop`

justinvdm:
@peterp ready for re-review.

peterp:
👍 , I really do think the action creators, and actions are very clear now.

PR #90, opened 2016-08-31
Fix chaining middleware to ignore actions dispatched from itself

justinvdm:


justinvdm:
@peterp ready for review

justinvdm:
Ok, now properly fixed and tested out in an emulator.

peterp:
👍 , after travis is happy, are ypu fixing this for when we may need it again?

justinvdm:
>  are ypu fixing this for when we may need it again?

@peterp I think we may still use this for api fetching.

PR #86, opened 2016-09-01
Connect up api loading

justinvdm:
#81 adds the things we need for loading from the api and showing a loading view. This PR is for using those things.


justinvdm:
@peterp ready for review!

src/constants/entry.js
```diff
@@ -1,10 +1,12 @@
-const LOAD_REQUEST = 'LOAD_REQUEST';
-const LOAD_SUCCESS = 'LOAD_SUCCESS';
-const LOAD_FAILURE = 'LOAD_FAILURE';
+const NEW_USER_ENTER = 'NEW_USER_ENTER';
+const EXISTING_USER_ENTER_REQUEST = 'EXISTING_USER_ENTER_REQUEST';
+const EXISTING_USER_ENTER_SUCCESS = 'EXISTING_USER_ENTER_SUCCESS';
+const EXISTING_USER_ENTER_FAILURE = 'EXISTING_USER_ENTER_FAILURE';
```
justinvdm: I'm not very confident on the naming here. `NEW_USER_ENTER` is dispatched for users with incomplete profiles.

`EXISTING_USER_ENTER_{REQUEST,SUCCESS}` is dispatched for the process of an existing user entering, which at the moment only involves loading. I went back and forth on naming the existing user routes `LOAD_{REQUEST_RESPONSE}`, they make sense in a way, but aren't really describing what is happening in the application at this point (the thing happening in the application at this point is the user is entering the entry point of the application).

@peterp would appreciate some ideas on names here :)

src/containers/WelcomeContainer.js
```diff
@@ -1,13 +1,12 @@
...
 import { startProfile } from 'src/actions/onboarding';
+import { getContext } from 'src/stores/helpers';
 
-import { getProfileId } from 'src/reducers/auth';
-import { getProfile } from 'src/reducers/profile';
 
 export default connect(
   state => ({
-    name: getProfile(state, getProfileId(state)).firstName,
+    name: getContext(state).profile.firstName,
```
peterp: I'm a bit concerned about using `getContext` here, instead of a selector. I think there's a benefit to having a "selector-like-api" as a common way to retrieving things from a reducer.

src/constants/entry.js
peterp: This is tough; we're redirecting the user after they've authenticated `login -> authenticated -> { new place }`

Just off the top of my head: `NEW_USER_ENTER` and  `EXISTING_USER_ENTER` should be part of `auth`

The `LOAD` step should be somewhere else; perhaps in `sync` ?

peterp:
👍

src/constants/entry.js
justinvdm: The distinction between login and entry is done since entry into the app happens in different ways:
  - after login for existing users
  - after onboarding for new users.

I like load being a separate action somewhere else. I'm going to try something along those lines.

src/containers/WelcomeContainer.js
justinvdm: I'm definitely for helpers for retrieving stuff from state. We could look things up off the state directly, but it would be nicer and DRYer to have helpers in `stores/helpers` to access things more easily. For example, to get a category's activities, theres a few small hoops to jump through that could be wrapped up in a `getCategoryActivities()` helper. Helpers also means we are protected against the state's naming and structure changing, since only the helpers would need to change for that case.

I don't really see much of a difference in value here between using `getContext()` and a `getProfile()` helper. Decided to use `getContext()` here since we already get the profile for context for our `enter()` action creator, but I'll use a `getProfile()` helper here if you'd prefer.

peterp: When I review the code I'm usually trying to make sure that it would make sense to someone who hasn't worked with the code before, as an example: 

If I gave them this container to compare against another they may be confused by the usage of `getContext()` here, which may lead them to dive into `getContext()` unnecessarily, and additionally assume that it's safe/ a pattern to add additional properties to `getContext()`

peterp: 👍 on the helpers

src/constants/entry.js
peterp: Cool, like you mentioned in Slack. Perhaps we should just do a general clean up a bit later... when we've got time.

justinvdm:
@peterp ready for re-review.

peterp:
👍

PR #93, opened 2016-09-05
Fix loading when existing users log in

justinvdm:
There were two issues:
  - the chaining middleware didn't allow target actions dispatched from it to re-enter the middleware as source actions
  - we were accidentally dispatching `ENTER_NEW_USER` for both new and existing users


justinvdm:
@peterp ready for review

peterp:
👍

PR #95, opened 2016-09-05
NavTabBar component

justinvdm:


justinvdm:
@peterp ready for review

justinvdm:
screen shot 2016-09-05 at 5 02 52 pm

src/components/NavTabBar/styles.js
```diff
@@ -0,0 +1,38 @@
...
+  },
+  tab: {
+    flex: 1,
+    alignItems: 'center',
+    justifyContent: 'center',
+  },
+  tabIcon: {
+  },
+  tabIconIsActive: {
+    tintColor: colors.NAV_TAB_ACTIVE,
```
justinvdm: This means we lose the subtle shadow in the icons for the active icon. Can't seem to be able to export the selected icons from the designs. It doesn't seem like there's a noticeable difference though, seems low priority.

src/constants/navigation.js
```diff
@@ -4,9 +4,21 @@ const NAVIGATE_FORWARD_REQUEST = 'NAVIGATE_FORWARD_REQUEST';
 
 const SCREEN_DISMISS = 'SCREEN_DISMISS';
 
+const NAV_TAB_ACTIVITIES = 'NAV_TAB_ACTIVITIES';
+const NAV_TAB_CHAT = 'NAV_TAB_CHAT';
+const NAV_TAB_COMMUNITY = 'NAV_TAB_COMMUNITY';
+const NAV_TAB_JOURNEY = 'NAV_TAB_JOURNEY';
+const NAV_TAB_SCHEDULED_CALLS = 'NAV_TAB_SCHEDULED_CALLS';
```
justinvdm: Not convinced this is the right place for these constants. Put them here because they are technically navigation related. We are going to need to access them from the nav reducer (or some reducer at the very least), this is why it isn't in the component's module.

peterp:
👍

PR #97, opened 2016-09-06
Onboarding with `Stepper`

peterp:


peterp:
@justinvdm ready for review. Sorry for the size of this.

src/actions/onboarding.js
```diff
@@ -1,25 +1,72 @@
...
+
+
+export {
+  stepBack,
+  stepForward,
+  save,
+  changeProfile,
+  changeProfilePicture,
+  chooseProfilePicture,
+};
```
justinvdm: Really like these action creators 🍰

src/components/Stepper/index.js
```diff
@@ -22,7 +22,7 @@ const Stepper = ({
   if (!hideProgress) {
     progressBar = (
       <ProgressBar
-        completed={ navigationState.index + 1 / React.Children.count(children) }
+        completed={ (navigationState.index + 1) * 1.0 / React.Children.count(children) * 1.0 }
```
justinvdm: Could you explain this change?

src/constants/onboarding.js
```diff
@@ -1,18 +1,40 @@
-const ONBOARDING_START_PROFILE = 'ONBOARDING_START_PROFILE';
 const ONBOARDING_CHOOSE_PROFILE_PICTURE = 'ONBOARDING_CHOOSE_PROFILE_PICTURE';
+const ONBOARDING_CHANGE_PROFILE_PICTURE = 'ONBOARDING_CHANGE_PROFILE_PICTURE';
+const ONBOARDING_CHANGE_PROFILE = 'ONBOARDING_CHANGE_PROFILE';
 
-const ONBOARDING_UPDATE_PROFILE_PICTURE = 'ONBOARDING_UPDATE_PROFILE_PICTURE';
-const ONBOARDING_UPDATE_PROFILE = 'ONBOARDING_UPDATE_PROFILE';
+const ONBOARDING_STEP_BACK_REQUEST = 'ONBOARDING_BACK_REQUEST';
+const ONBOARDING_STEP_FORWARD_REQUEST = 'ONBOARDING_NEXT_REQUEST';
```
justinvdm: I don't think this is too important considering we are planning on cleaning up our action names in general, but since we're adding new ones, I'm wondering if we should use this pattern:
  - for _command_ actions (actions triggered programatically), use `<NOUN>_<VERB>_{REQUEST,SUCCESS,FAILURE}`
  - for actions _describing something that happened_ (not something triggered by the programmer): `<NOUN_VERB>`

I think my original reasoning for wanting to use `REQUEST` for the second type of action might have been wrong. I was thinking of `REQUEST` as "the user is requesting for this action to happen in the application", rather than "we've given a command, this action is for describing it is pending". Not really against us using `REQUEST` in both cases though.

Happy for this to stay as is and for this to change as part of the general clean up.

src/reducers/routes/onboarding.js
```diff
@@ -1,28 +1,23 @@
...
 } from 'src/navigationHelpers';
 
 
 export default (state = createStack([
-  createRoute(routes.ROUTE_ONBOARDING_WELCOME),
+  createRoute(routes.ROUTE_ONBOARDING),
 ]), action) => {
   switch (action.type) {
-    case onboarding.ONBOARDING_START_PROFILE:
-      return pushList(state, routes.ONBOARDING_STEPS.map(unary(createRoute)));
```
justinvdm: Glad this is gone :)

src/views/ProfilePicture/__tests__/ProfilePicture-tests.js
```diff
@@ -0,0 +1,33 @@
...
+
+describe('ProfilePicture', () => {
+  const createComponent = (props = {}) => (
+    <Provider store={fakeStore}>
+      <ProfilePicture
+        profilePicture="path/to/image.png"
+        onChoosePhotoPress={noop}
+        {...props}
+      />
+    </Provider>
```
justinvdm: Could you remind me why we need `Provider` here (and for the other views)?

src/views/OnboardingSaving/index.js
```diff
@@ -0,0 +1,32 @@
...
+    const {
+      id,
+      profile,
+    } = this.props;
+
+    this.props.save({
+      id,
+      profile,
+    });
+  }
```
justinvdm: Could you explain why we need these? I would've thought we could have just simply wrapped `OverlayLoading`.

src/views/FormStep/index.js
```diff
@@ -0,0 +1,39 @@
...
+FormStep.propTypes = {
+  title: PropTypes.string.isRequired,
+  children: PropTypes.any.isRequired,
+  paginationDisabled: PropTypes.bool,
+  onBackPress: PropTypes.func.isRequired,
+  onNextPress: PropTypes.func.isRequired,
+};
+
+
+export default FormStep;
```
justinvdm: Should this maybe be a component rather than a view?

src/components/Stepper/index.js
peterp: I needed the results to be floats; I saw inconsistent results when I didn't do this... Not exactly sure why.

justinvdm:
Left a few minor comments. Also looks like there are merge conflicts with develop we should fix.
 
Really like how much `Stepper`, `FormStep` and `onboardingContainer()` simplify things, nice and coherent :)

src/views/ProfilePicture/__tests__/ProfilePicture-tests.js
peterp: That's because `<FormStep><ProfilePicture /></FormStep>` is <OnboardingFormStepContainer>; which connects the `stepForward()/ stepBack()` actions for the `OnboardingStepper`

src/components/Stepper/index.js
justinvdm: I've never really groked the weird edge cases with js numbers, but I'm surprised this has any affect on these numbers (regardless of whether `1` or `1.0` was used), seems like a no-op either way. Just wondering if the actual reason for the inconsistent results is still lurking about.

src/views/FormStep/index.js
justinvdm: Happy for us to do this later.

src/views/OnboardingSaving/index.js
peterp: I wanted to dispatch the save action when this view is shown. I don't, yet, have a way to determine when the stepper is on the last screen: I would like to improve the stepper by making it to dispatch an action `ONBOARDING_STEPPER_LAST/ DONE` that can be chained to call `save()`

and additionally adding `ONBOARING_STEPPER_INIT` just for good measure.

I didn't have the bandwidth to add that in now.

src/constants/onboarding.js
peterp: Ok, agreed, it's a bit legacy. 👍

src/views/ProfilePicture/__tests__/ProfilePicture-tests.js
justinvdm: Ah, I see. It would be cool if we didn't need to mix containers and views that way, like if there was a way we could pass in the form step props to a generic `FormStep` inside of `ProfilePicture` (for e.g. via some `stepProps` object or something), and have the `step{Forward,Back}` bindings for `FormStep` be passed down from the container for `ProfilePicture`.

If it isn't possible then that's cool. If it is, happy for that to happen in a later PR.

src/components/Stepper/index.js
peterp: I just changed/ tested it; it's not inconsistent... I think it may have been a lack of focus on my part.

src/views/ProfilePicture/__tests__/ProfilePicture-tests.js
peterp: Ok, I think that would clean this up nicely. Happy to try it; but also in a later PR.

src/views/OnboardingSaving/index.js
justinvdm: Hmm, I could be missing some use cases, but not really convinced we'd need `INIT`. For `INIT`, any logic that should take place in the reducer at the start of the stepper should already be doable as part of the logic for how the stepper appears in the first place.

Dispatching a `DONE` action when the user presses done/next on the last step, and chaining that to `save()` sounds awesome :) +1 on that approach, though happy with this approach for this PR.

src/views/ProfilePicture/__tests__/ProfilePicture-tests.js
peterp: Created #101

justinvdm:
:+1:!

src/views/FormStep/index.js
peterp: Created #102

src/views/OnboardingSaving/index.js
peterp: Created #103

PR #106, opened 2016-09-07
Better navigation approach

justinvdm:
Our current nav approach doesn't fit well with `NavigationExperimental`'s card-stack model, and one side effect of this is weird transitions when switching between nav tabs.

This ticket changes navigation to an approach that looks something like:

```
top level card stack
  \ landing
  \ onboarding
  \ tabbed navigation
    \ journey tab's card stack
    \ activities tab's card stack
```


justinvdm:
@peterp ready for review.

I haven't included handling for modals. I think it would be better to try that out when we add our first modal view, I have some ideas though.

src/app.js
```diff
@@ -5,12 +5,12 @@ import { Provider } from 'react-redux';
...
 const store = configureStore();
 
-import RootNavigation from 'src/containers/NavigationStackContainer';
+import TopNavigationContainer from 'src/containers/TopNavigationContainer';
 
 const App = function App() {
   return (
     <Provider store={store}>
-      <RootNavigation />
+      <TopNavigationContainer />
```
peterp: How do you feel about `BaseNavigation/Container` ?

peterp: Actually, scrap that, was a passing thought 💃

peterp:
👍  🍰

PR #108, opened 2016-09-08
Category List fallback colors

peterp:
@justinvdm ready for review

I originally wanted to generate a gradient of colors based on the amount of categories within the list; but I think that may be over engineering an unlikely scenario.


src/views/CategoryList/index.js
```diff
@@ -16,13 +17,18 @@ const CategoryList = ({
...
         <TouchableNativeFeedback
           key={id}
           categoryId={id}
           onPress={() => onCategoryPress(id)}
         >
-          <View style={[styles.category, { backgroundColor: color }]}>
+          <View
+            style={[
+              styles.category,
+              { backgroundColor: color || colors.CATEGORY_LIST_GRADIENT[index] },
```
justinvdm: Could we mod this index by the scale length, to ensure we have a color even we have more categories than colours? Maybe worth moving all that to a helper in this module.

src/helpers.js
```diff
@@ -1,6 +1,21 @@
+import convert from 'color-convert';
 import { find, isUndefined, omitBy, isNull } from 'lodash';
 
 
+export const makeGradient = (start, end, length) => {
+  const s = convert.hex.rgb(start);
+  const e = convert.hex.rgb(end);
+
+  return Array.from({ length }, (v, i) => {
```
justinvdm: I'm wondering how robust using `Array.from()` and faking an array as a way of creating a range is, but can't think of why it wouldn't work. Using lodash's `range()` seems safer, but not really important enough to hold up this PR.

justinvdm:
:+1:wtih 🍰 !

src/helpers.js
peterp: I sort of have a love/hate relationship with lodash. I love the functionality; but I feel it's a stop-gap for features that the JS should support... Perhaps I should just pretend that it's part of the language. :)

justinvdm: > Perhaps I should just pretend that it's part of the language. :)

Sort of what I do :) I'm probably awakening trolls and angering language gods by saying this, but I kind of like that js doesn't really have a standard lib. Just providing the core things and pushing everything else to user land sidesteps lots of scope creep, leverages all the trial and error happening in user land for surfacing library apis that work best for most use cases, and means less things that different js implementations could do differently.

PR #112, opened 2016-09-09
Activity section fixes

justinvdm:
- move category color fallbacks to api
- fix broken load() api method
- fix bad action creator import for `ActivityContainer`
- update entities on load success

PR #111, opened 2016-09-09
Handle loading failure

justinvdm:

PR #113, opened 2016-09-09
Fix back button pressing for activity screens

justinvdm:


justinvdm:
@peterp depends on #111 and #112, otherwise ready for review.

peterp:
👍

PR #115, opened 2016-09-09
Journey Screen

peterp:
@justinvdm Ready for review

screen shot 2016-09-15 at 03 05 54

screen shot 2016-09-15 at 03 05 59


src/views/Journey/index.js
```diff
@@ -0,0 +1,90 @@
...
+  onGetStartedPress,
+}) => (
+  <BaseView>
+    <Header style={styles.header}>
+      <Text style={styles.nextCall}>NEXT CALL</Text>
+      <Link
+        style={[styles.date, nextScheduledCallDate && styles.hasDate]}
+        onPress={onNextScheduledCallPress}
+      >
+        {nextScheduledCallDate || 'Schedule your first call'}
```
justinvdm: Wondering if we should format the date here with `moment`.

src/views/Journey/index.js
```diff
@@ -0,0 +1,90 @@
...
+
+    <View style={styles.eventsContainer}>
+      <Image
+        style={styles.eventsBackground}
+        source={images.JOURNEY_BG}
+      >
+        <ScrollView>
+          <EventGreeting
+            onGetStartedPress={onGetStartedPress}
+          />
```
justinvdm: Will we maybe need to use a prop to decide whether we should display this or to display events? Sounds like a change for a different PR though.

src/views/Journey/index.js
peterp: @justinvdm I would like to actually have scheduled call in the app before I do that;

PR #116, opened 2016-09-14
Use orange active state icons in NavTabBar

justinvdm:


justinvdm:
@peterp ready for review

PR #117, opened 2016-09-14
Fix NavTabBar to not overlap with view

justinvdm:


justinvdm:
@peterp ready for review

PR #118, opened 2016-09-16
Connect up push notifications

justinvdm:
We need to use [react-native-fcm](https://github.com/evollu/react-native-fcm), update the api using the auth token we get from it, and dispatch actions when we receive notifications.


justinvdm:
@peterp ready for review

justinvdm:
_note:_ still needs to be tested out, waiting on a download to get GCM working with genymotion.

PR #120, opened 2016-09-17
Add StartCall view

justinvdm:


justinvdm:
screen shot 2016-09-17 at 6 09 38 pm
screen shot 2016-09-17 at 6 18 56 pm
screen shot 2016-09-17 at 6 43 56 pm

justinvdm:
@peterp ready for review

src/views/StartCall/styles.js
```diff
@@ -0,0 +1,30 @@
...
+    flex: -1,
+    width: 328,
+    height: 288,
+    resizeMode: 'contain',
+  },
+  contentContainer: {
+    flex: 1,
+    paddingLeft: 24,
+    paddingRight: 24,
+  },
```
justinvdm: After lots of trying out of different things, this was the best approach I could find for fitting the image inside the screen's dimensions, while leaving enough space for the content and footer:

Set the image to its actual dimensions, then shrink it to fit in these constraints:
    - `imageWidth` is larger than `screenWidth - (paddingLeft + paddingRight)`
    - `imageHeight + paddingBottom` is larger than `3/4` of the body (the content takes up `1/4`).

This should keep the image at the same size regardless of the amount of content without any awkward truncation, and should let us fit the image sensibly for varying screen widths.

peterp: `resizeMode: 'contain'`, should do exactly that, if you set the `height: null` and `width: null` for the image.

src/views/StartCall/index.js
```diff
@@ -0,0 +1,76 @@
...
+            {activity.title}
+          </Text>
+        }
+      </View>
+    </View>
+
+    <View style={styles.footer}>
+      <Button
+        uid="activate"
+        layout={Button.layouts.stretch}
```
peterp: I wasn't aware of the `layoutSelf` property! Nice! I actually think that this should be part of the default layout.

src/views/StartCall/styles.js
```diff
@@ -0,0 +1,33 @@
+import { StyleSheet } from 'react-native';
+
+
+export default StyleSheet.create({
+  body: {
+    flex: 1,
+  },
+  imageContainer: {
+    flex: 3,
```
peterp: I think it's easier to think about the flex properties when they're decimals, thoughts?

src/views/StartCall/styles.js
```diff
@@ -0,0 +1,33 @@
...
+    alignItems: 'center',
+    paddingLeft: 16,
+    paddingRight: 16,
+    paddingBottom: 29,
+  },
+  image: {
+    flex: -1,
```
peterp: `
width: null,
height: null,
resizeMode: 'contain',
`

and removing `flex: -1` may give us the results we need.

src/views/StartCall/index.js
```diff
@@ -0,0 +1,76 @@
...
+const StartCall = ({
+  callTime,
+  activity,
+  onDismissPress,
+  onActivatePress,
+}) => (
+  <BaseView>
+    <Header>
+      <HeaderIcon
+        uid="dismiss"
```
peterp: Is `id` a reserved keyword?

justinvdm: I don't think so, but avoiding it in case we at some point want to port these components to the web (unlikely but not impossible).

src/views/StartCall/styles.js
justinvdm: +1, done.

justinvdm:
Ready for re-review, I think.

PR #123, opened 2016-09-19
Journey Event Types

peterp:


src/stores/__tests__/helpers-test.js
```diff
@@ -133,4 +142,59 @@ describe('helpers', () => {
       ]);
     });
   });
+
+  describe('mapScheduledCallEvent', () => {
```
peterp: @justinvdm can you let me know what you think of this direction?

scripts/helpers.js
```diff
@@ -57,13 +58,37 @@ export const fakeActivity = data => ({
   ...data,
 });
 
+export const fakeScheduledCall = data => ({
+  id: 1,
+  createdAt: '2016-09-16T11:27:25.454520Z',
+  callTime: '2016-09-16T11:27:14Z',
```
justinvdm: I'm wondering whether the api should do the serializing and parsing necessary for the entities we work with to use milliseconds (e.g. `+new Date()`). That way we don't have to know to parse and serialize to a given format all over the place. Not something to add to this PR, but if we go that route, we should use milliseconds here. Thoughts?

src/containers/MilestoneEventContainer.js
```diff
@@ -0,0 +1,28 @@
...
+import { getCategory } from 'src/stores/helpers';
+
+
+const mapStateToProps = (state, {
+  objectId: categoryId,
+  eventType: type,
+  occuredAt: date,
+}) => {
+  const {
+    image,
```
justinvdm: The designs seem to suggest we should be using a category icon for the milestone event. We don't seem to get that from the api though. Maybe we should raise this and use this so long?

justinvdm:
I think we need tests and screenshots for the components.

justinvdm:

src/containers/MilestoneEventContainer.js
peterp: Aaah, interesting. I assumed this was the icon; 👍

scripts/helpers.js
peterp: 👍

PR #128, opened 2016-09-19
Get FCM working

justinvdm:


package.json
```diff
@@ -19,8 +19,8 @@
     "moment": "^2.12.0",
     "normalizr": "^2.2.1",
     "promise-each-concurrency": "^1.1.0",
-    "react": "~15.2.1",
-    "react-native": "^0.31.0",
+    "react": "^15.3.1",
+    "react-native": "^0.33.0",
```
justinvdm: `react-native-fcm@2` requires `react-native@~0.33`. We could use an earlier version of `react-native-fcm` now and upgrade later once react-native `0.34` is out and we don't have to workaround things as much in tests (although we are already working around things quite a bit at the moment), but then we'd have to go through the trouble of upgrading to `react-native-fcm@2` (since it would be an upgrade to a new major version, we'd probably need to deal with backward incompatibilities). Thought we might as well bite the bullet and get things working with newer react and react-native.

justinvdm:
@peterp ready for review

Still busy getting receiving from FCM working, but that is taking a while, and since we seem to be getting tokens using `react-native-fcm` now, thought it would be worth making a PR for what we have so far.

peterp:
👍

PR #131, opened 2016-09-20
Start call routing and navigation

justinvdm:

PR #134, opened 2016-09-21
Add routing for call completed notification

justinvdm:


justinvdm:
@peterp depends on #131, otherwise ready for review

PR #135, opened 2016-09-21
Pass activity when creating call

justinvdm:


justinvdm:
@peterp ready for review

PR #138, opened 2016-09-21
Support navigating to start call screen from journey screen

justinvdm:


justinvdm:
@peterp ready for review

PR #148, opened 2016-09-26
Callback when a tab is switched

peterp:


peterp:
@justinvdm ready for review

I'm not sure why the 3 top commits are here... as they should've been merged into develop, master and release when the tag worked.

justinvdm:
👍

PR #149, opened 2016-09-27
Call Notes API

peterp:
@justinvdm ready for review

### DIFF
--- a/scripts/jest/environment.js
+++ b/scripts/jest/environment.js
@@ -60,3 +60,6 @@ global.render = (...args) => {
 
 
 moment.tz.setDefault('Africa/Johannesburg');
+
+// guaranteed to have been now at some point in time
+Date.now = () => 1474966266635;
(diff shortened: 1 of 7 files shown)


scripts/jest/environment.js
```diff
@@ -60,3 +60,6 @@ global.render = (...args) => {
 
 
 moment.tz.setDefault('Africa/Johannesburg');
+
+// guaranteed to have been now at some point in time
+Date.now = () => 1474966266635;
```
justinvdm: Ah, created #151 for this, maybe best to merge that first?

src/api/callNotes.js
```diff
@@ -0,0 +1,38 @@
...
+  auth,
+});
+
+
+export const patchCallNote = (id, data, auth) => request({
+  url: `/call_note/${id}/`,
+  method: 'PATCH',
+  data,
+  auth,
+});
```
justinvdm: If we need to update our entities, for all these methods, I think we need to add a schema prop to each of them.

justinvdm:
Looks good :) just the one small thing.

justinvdm:
👍 wtih

PR #153, opened 2016-09-27
Allow choosing past scheduled calls

justinvdm:


justinvdm:
@peterp ready for review

PR #163, opened 2016-09-28
Add back buttons as last child in headers to ensure they receive presses

justinvdm:


justinvdm:
@peterp ready for review

PR #164, opened 2016-09-29
Replace-or-push routes instead of replace, in case the old routes have been dismissed

justinvdm:


justinvdm:
@peterp ready for review

PR #166, opened 2016-09-29
Schedule Call AM/PM clock

peterp:
screen shot 2016-09-29 at 16 23 39

screen shot 2016-09-29 at 16 24 46

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "app",
-  "version": "0.1.4",
+  "version": "0.1.5",
   "private": true,
   "scripts": {
     "start": "node node_modules/react-native/local-cli/cli.js start",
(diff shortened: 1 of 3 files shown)


peterp:
@justinvdm ready for review

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "app",
-  "version": "0.1.4",
+  "version": "0.1.5",
```
justinvdm: I'd rather we did version bumping using a release branch, or is this a commit that somehow managed to appear in the diff but isn't for this PR?

justinvdm:
👍 !

package.json
peterp: Yup. I don't think `git flow` pushes the develop branch after it's merged. 🍤

PR #173, opened 2016-09-30
Fix image urls

justinvdm:
At the moment we assume it is prefixed with `/mentor`


justinvdm:
@peterp ready for review

PR #174, opened 2016-09-30
Handle case where length === 1 for makeGradient

justinvdm:


justinvdm:
@peterp ready for review

I noticed it would use `#000` for a length of `1`.

PR #180, opened 2016-09-30
Navigator Local State to Redux

peterp:

PR #183, opened 2016-09-30
Display next scheduled call date

peterp:
@justinvdm depends on #182, #180 -- but otherwise ready for review. (last two commits.)


justinvdm:
:+1:wtih

PR #185, opened 2016-10-01
Fixed schedule detail time formatting

peterp:
12 hour clock, with AM/ PM instead of 24 hour clock.

@justinvdm ready for review


justinvdm:
:+1:

PR #188, opened 2016-10-03
using scheduled call call time instead of event date

peterp:


peterp:
@justinvdm ready for review

justinvdm:
👍

PR #189, opened 2016-10-03
`getNextScheduledCall` should work if there arent any dates in the past

peterp:


peterp:
@justinvdm ready for review

justinvdm:
:+1:wtih

justinvdm:
actually, travis was happy before the test description change, :+1:

PR #192, opened 2016-10-03
Prevent user from choosing times in the past

justinvdm:


justinvdm:
/o\ was on wrong branch when creating PR, sorry for the noise.

PR #193, opened 2016-10-03
Prevent user from choosing times in the past

justinvdm:


justinvdm:
screen shot 2016-10-03 at 4 20 58 pm

justinvdm:
@peterp ready for review

PR #199, opened 2016-10-05
Upgrade Jest to 16.0.0

peterp:


peterp:
@justinvdm ready for review

package.json
```diff
@@ -37,7 +37,7 @@
     "eslint-plugin-react": "^4.2.1",
     "immediate-promise": "^1.1.0",
     "isomorphic-fetch": "^2.2.1",
-    "jest": "^15.1.1",
+    "jest": "16.0.0",
```
justinvdm: Should we maybe use `~16.0.0` so we still get patch releases?

justinvdm:
One small comment, otherwise :+1:wtih.

PR #200, opened 2016-10-06
Auto-populate calendar dates when adding scheduled call

justinvdm:


justinvdm:
@peterp ready for review

PR #203, opened 2016-10-07
Support removing activities from scheduled calls

justinvdm:

PR #204, opened 2016-10-07
Prevent past scheduled calls from being resheduled in past

justinvdm:
@peterp depends on #203, otherwise ready for review.


justinvdm:
screen shot 2016-10-07 at 5 43 35 pm
screen shot 2016-10-07 at 5 43 44 pm

PR #206, opened 2016-10-10
Call Notes

peterp:
screen shot 2016-10-10 at 13 13 16

screen shot 2016-10-10 at 13 13 42

screen shot 2016-10-10 at 13 31 16

screen shot 2016-10-10 at 13 31 09


peterp:
@justinvdm ready for review

src/components/RadioList/index.js
```diff
@@ -0,0 +1,60 @@
...
+import { Text } from 'src/components';
+import images from 'src/constants/images';
+import styles from './styles';
+
+
+class RadioList extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      selectedIndex: this.props.initialSelectedIndex || void 0,
```
justinvdm: If the given `initialSelectedIndex` was `0`, this would become `undefined`. Guessing we want to check for undefined instead (maybe using lodash's `isUndefined` for consistency)?

src/views/CallNoteSteps/index.js
```diff
@@ -0,0 +1,207 @@
...
+
+
+const CallQuality = ({
+  callQuality = '',
+  onChangeText,
+}) => {
+  const items = [
+    'Excellent',
+    'Ok',
+    'Couldn\'t hear',
```
justinvdm: Bikeshed: could we use `"` instead?

justinvdm:
Looks good :) left minor comments.

PR #210, opened 2016-10-11
Logout action and reducing

justinvdm:
This adds logic for logging the user out of the app.


justinvdm:
Ready for review.

Was going back and forth on clearing out all state on logout. Turns out this isn't really easy to do at the moment, since we need to keep all non-nav state around until the nav transitions are done. We can look into doing this, but I think an argument can be made for _not_ clearing all state on logout. For example, there may later on add some initial state loaded when the app starts (rather than on login), or it may be better for the user's current nav tab to be preserved for when the user logs back in.

smn:
:+1:

PR #211, opened 2016-10-11
Profile sign out skeleton

justinvdm:
This is for the basic logic for navigating to and from the profile settings screen and for supporting signing out on this screen.

Ready for review


justinvdm:
Note that this isn't the PR where we aim on matching the designs perfectly for this screen, that is for a later PR (functionality more important for this PR).

screen shot 2016-10-11 at 6 28 29 pm

justinvdm:
Note for review: this PR depends on #210.

smn:
:+1: but what are the .snap files for?

justinvdm:
> 👍 but what are the .snap files for?
https://facebook.github.io/jest/blog/2016/07/27/jest-14.html

We are using it for testing that components are being rendered in a way that we have confirmed is correct (thus the screenshots, to show that we have in fact confirmed it is being rendered correctly). For testing that the callback properties we give a component are called when UI events happen, we use [enzyme](https://github.com/airbnb/enzyme) to simulate the UI events.

PR #213, opened 2016-10-12
Temporarily use placeholder profile icon on Journey screen

justinvdm:
Once we are able to get scaled-down versions of the profile image from the API we can use that instead. In the meantime, we need a placeholder profile icon.

Ready for review


justinvdm:
screen shot 2016-10-12 at 1 41 57 pm

justinvdm:
Note for review: depends on #211.

smn:
:+1: on the 1 commit added after #211

PR #215, opened 2016-10-13
Tie up loose ends for call notes

justinvdm:
There are a few loose ends in call notes at the moment:
  - one should be able to view call notes associated to an activity from the activity detail screen
  - we are sending different fields than what the api is expecting when creating call notes

Ready for review


justinvdm:
screen shot 2016-10-13 at 8 05 59 pm
screen shot 2016-10-13 at 6 43 20 pm



screen shot 2016-10-13 at 8 01 02 pm

smn:
This looks good to me but I don't have the full context on the call notes to be able to review this well.

PR #216, opened 2016-10-14
Support viewing all call notes

justinvdm:
At the moment we support viewing call notes related to an activity. We need a way of allowing the user to view _all_ call notes.


justinvdm:
Depends on #215, otherwise ready for review.

justinvdm:
screen shot 2016-10-14 at 2 11 56 pm

smn:
:+1: on the 3 commits added after #215

PR #217, opened 2016-10-17
Start scheduled call from Journey top bar

justinvdm:
At the moment, the only way for the user to reach the start call screen for a scheduled call is via the 1 min call reminder. The plan is to change the top bar on the Journey screen so that tapping on it takes the user to the start call screen for a scheduled call if the current time is less than 1h before or after the scheduled call's time.

Depends on #213, otherwise ready for review.


justinvdm:
screen shot 2016-10-17 at 4 35 23 pm
screen shot 2016-10-17 at 4 35 54 pm
screen shot 2016-10-17 at 4 36 35 pm

justinvdm:

smn:
:+1: on this from me but I'm still not sure I understand how the testing works with the .snap files.

justinvdm:
> 👍 on this from me but I'm still not sure I understand how the testing works with the .snap files.

Thanks for review :)

Will try explain:

When we run a test that uses `expect(...).toMatchSnapshot()` for the first time, it will save a `.snap` file containing a snapshot of the rendered component's properties (and all its childrens' properties).

The next time the test is run, instead of creating a `.snap` file, it will compare the rendered component's properties to the properties in the snapshot, and fail the test if they do not match.

We can then explicitly update the `.snap` files if the changes are valid (and if it makes sense to, we usually add a screenshot to the PR confirming that we've checked that the change is valid).

PR #218, opened 2016-10-18
Check for journey screen's top bar updates every min

justinvdm:
As of #217, the journey screen's top bar renders a next scheduled call reminder, but switches to show "start call now" (where tapping would allow the user to start the call) if the current time is 1 hour or less before or after the most recent sheduled call time.

We need a way for this bar to update when it should switch to its "start call now" mode.

The plan is to dispatch a tick action every minute, and use this for updating the Journey view. Problem is, react-redux won't simply re-render if an action has been dispatched, since the app's state hasn't changed at all (it seems to do a shallow comparison between the new and old state).

These are the approaches I could think of to work around this:
  1. in a reducer, record the current time each tick and add it to the state
  2. move the logic for computing the most recent scheduled call, as well as whether a call should be started, from the Journey view's container to a reducer

(2) complicates our reducer setup, since its reducer would need access to the entire state to use our store helpers, and since technically this is computable state. It _feels_ like this logic belongs more in a container, so I'm going with approach (1) for now.

Ready for review.

### DIFF
--- a/src/actions/__tests__/tick-test.js
+++ b/src/actions/__tests__/tick-test.js
@@ -1,0 +1,28 @@
+import tick from 'src/actions/tick';
+import * as constants from 'src/constants/tick';
+
+jest.useFakeTimers();
+
+
+describe('actions/tick', () => {
+  it('should dispatch a tick action each interval', () => {
+    const dispatch = jest.fn();
+    tick()(dispatch);
+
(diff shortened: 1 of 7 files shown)


justinvdm:

smn:
:+1:

PR #221, opened 2016-10-20
Support both production and qa apks

justinvdm:
At the moment we only support production. Once we're live, we will need separate apks using separate configs.


justinvdm:
@smn ready for review

smn:
good idea but the travis build is failing

justinvdm:
hrm, for some reason importing `react-native-config` makes tests timeout. I've changed the config stuff so it is only imported if we are in qa or prod, since we don't need it in tests and dev anyways. Actually running the app seems to work fine, so I'm guessing it has something to do with the fact that tests are run in a very different environment to the actual app.

smn:
:+1: also it's late.

PR #222, opened 2016-10-20
Get the basics set up for sentry

justinvdm:
@smn depends on #221, otherwise ready for review.


justinvdm:
Note for review: We still need to upload the bundle and sourcemaps for the bundle to sentry as release artefacts. This PR is just for the basics.

smn:
I'm assuming this would mean we ship the apk with the sentry url baked in?

smn:
:+1:

justinvdm:
> I'm assuming this would mean we ship the apk with the sentry url baked in?

@smn yep, anything I'm missing that could make that a bad idea?

I guess it means a sentry url change would require users to download a new release. Is that case likely enough that we should avoid this approach?

smn:
no I think that's fine

PR #223, opened 2016-10-21
Actually get sentry working

justinvdm:
@smn ready for review

I don't think sourcemaps are working, but it is difficult to test without actually building the apk, which takes a while. It _might_ be working, but if it isn't, I think figuring out why is likely to take more time than we have at this point.


smn:
:+1: as I suspect the build failed because of global DNS errors?

PR #224, opened 2016-10-24
Move activity nav logic to top nav reducer

justinvdm:


justinvdm:
(I think the travis failures are because of an irrelevant dns-related issue)

smn:
nope, build failures due to openssl decrypts failing.

justinvdm:
> nope, build failures due to openssl decrypts failing.

ya, sorry, was looking at wrong build :/ I think the failure might be the result of me accidentally not checking in the latest `secrets.tar.enc` after running `travis encrypt-file ...` (so they key was updated on travis-ci.org but was being used on the old file), though I don't remember anything like that happening, so not too sure.

I've updated done a new `travis encrypt-file ...` and updated develop with the new `secrets.tar.enc`, will see if that fixes things.

justinvdm:
Build passing now, ready for review.

smn:
:+1:

PR #231, opened 2016-10-26
Adjust styles on landing view to handle smaller screens better

justinvdm:


justinvdm:

Not exactly an amazing improvement, but slightly better than what we had for smaller screens. Not much more we can do without adding scrolling into the mix or getting a new image that covers better on smaller screens.

before:
screen shot 2016-10-26 at 6 59 00 pm

after:
screen shot 2016-10-26 at 7 02 06 pm
screen shot 2016-10-26 at 7 00 05 pm

Mitso:
@justinvdm 👍 
Just an idea, how does _vertical-align: top; margin: 0;_ render the image? To start the image from CTA top to bottom and remove the empty space at the top there.

justinvdm:
@Mitso could you take another look at this? I've changed things so that we can keep to the initial design if the screen is >480px. After gettings things working so that the image was vertically aligned to the top, it looked quite bad (on both screen sizes, it was hardly showing the faces in the illustration), so I think it is best keeping it centered as it is at the moment.

screen shot 2016-10-27 at 4 25 32 pm
screen shot 2016-10-27 at 4 25 44 pm

Mitso:
@justinvdm ook cool NP, look good 👍

PR #232, opened 2016-10-27
Resize category about image to match design change

justinvdm:


smn:
:+1: when travis

justinvdm:
Example image isn't the greatest example, but does its job of showing that we are allowing for 30px (relative to screen density) on either side:

screen shot 2016-10-27 at 8 58 07 pm
screen shot 2016-10-27 at 8 59 44 pm

PR #234, opened 2016-10-28
Clear activity data on sign out

justinvdm:
If we don't do this, the activities will stay around, even if they have been removed.

Note that if we land this, we need the api to never remove activities and articles, but to rather set them as archived or unpublished, and filter on this. Otherwise, we will not be able to get the data we need for other entities with relations to activities (e.g. call notes).


smn:
:+1: but do we need to make sure the backend work is done before merging this?

justinvdm:
@smn thanks for the review :)

The backend work is underway, so I'm going to merge since without it we will show activities that we shouldn't show anymore, and since #235 allowed us to handle non-existent entities more gracefully.

PR #235, opened 2016-10-28
Handle missing and hidden entitites better

justinvdm:
If we some entitity has a relation to some other entity, but for some reason that other entity doesn't exist anymore (for e.g. a django admin or cms change that wasnt handled properly), at the moment we just explode. The plan is to:
- filter out events with missing related entities
- not associate any activity with entities we draw UI for if their activity is missing

We should also be supporting hiding entitities: expect the api to do a 'soft delete' or 'soft unpublish' and hide entities with an `is_hidden` field to true.


justinvdm:

smn:
:+1:

PR #240, opened 2016-11-03
Fix android app configuration

justinvdm:
Turns out `com.app` isn't the greatest namespace if we want our app to be unique on the device and on google play store.

We also aren't using the version in our `package.json` to configure the actual version details used by the device and google play store.

@smn ready for review


android/app/build.gradle
```diff
@@ -89,11 +89,11 @@ android {
     buildToolsVersion "23.0.1"
 
     defaultConfig {
-        applicationId "com.app"
+        applicationId "com.praekelt.mentorship"
```
smn: org.praekelt

justinvdm: Thanks, for some reason this made me realise I still need to update the `google-services.json` files.

justinvdm:
@smn ready for re-review

PR #243, opened 2016-11-08
Fix Journey screen scrollable area

justinvdm:
The top area with the illustration and call button should scroll with the events below it


justinvdm:
In the process, I realised we need to support repeating patterns (we don't get this feature out of the box for android so a new component needed to be created for this), and updating the pattern image.

justinvdm:
@smn @Miso @projectmushroom ready for review

Mitso:
@justinvdm +1 on the PatternBackground.

PR #245, opened 2016-11-10
Fix transitions

justinvdm:
The navigation we get out of the box with react native doesn't give us much control for cases that aren't as simple as pushing on a route or popping off a route.

This affects us most when multiple routes are popped off at once (e.g. when logging out, or when choosing something via a series of screens). Our current navigation will transition out all the popped off routes in quick succession, which looks buggy and confusing. What would be better would be to have it appear as if the new "top of the stack" was pushed on (as if the user has just completed a cycle of screens and arrived back at the starting screen).

After getting lost in react native navigation and animation rabbit holes and trying a few things out, the simplest way to get what we want is to roll our own navigation transitioning. It turned out way less hacky and much simpler than trying to get the stock navigation components to do what we want.


justinvdm:
@smn @Mitso @projectmushroom reminder that this is ready for review

smn:
:+1:

justinvdm:
Thanks for review @smn. Getting seemingly unrelated travis failures that I'm struggling to reproduce locally, looking into it before landing.

justinvdm:
Finally, travis is fixed. It seems that the version range `~16.0.0` for jest was ending up giving us a different version on travis `16.0.0` or `16.0.1`, while on the two machines I tried it on locally it was resolving to the `16.0.2`. Changing the range to `^16.0.2` seems to have fixed it.

PR #247, opened 2016-11-18
Use programmatic call note values

justinvdm:
@codiebeulaine could you review this? Think you'd have the most context

### DIFF
--- a/src/constants/callNote.js
+++ b/src/constants/callNote.js
@@ -52,6 +52,31 @@ const EVENT_MOOD_IMAGES = fromPairs([
 ]);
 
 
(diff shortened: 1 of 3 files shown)


codiebeulaine:
@justinvdm 👍 wtih. How will this change things on the backend?

justinvdm:
@codie Nothing on the backend needs to change. What will be different though, is that instead of storing values like `Not at all` for the `activity_helpful` field or `Excellent` for the `call_quality` field, it will now be storing `0` for `activity_helpful` and `EXCELLENT` for `call_quality`. Important thing is that the copy can now change without altering what we are sending to the api.

codiebeulaine:
@justinvdm 👍

PR #248, opened 2016-11-18
Prevent duplicate call notes

justinvdm:
It seems we still have a duplicate call notes issue. This is still a workaround, but prevents duplicate call notes since this is a change at the action layer, rather than the view layer.

Fixing this properly involves quite a bit of refactoring and cleanup, of which #246 is the start, but we need this fixed sooner than that.


justinvdm:
@smn @Mitso @projectmushroom reminder that this is ready for review

codiebeulaine:
👍 🍰

PR #254, opened 2016-11-24
Remove dead code

justinvdm:

PR #255, opened 2016-11-24
Better api action tests

justinvdm:
At the moment, our tests for action creators that use the `apiAction()` helper aren't great.

Since this helper does all the logic for making api requests and dispatching the relevant actions, any action creators using this helper basically become configuration more than anything.

Since we still wanted to have these action creators tested, we ended up writing tests that basically looked identical to the implementation, which made for pretty useless tests (at the very least though, they were serving as some basic sort of regression tests).

This PR introduces a helper that actually checks that the correct actions were dispatched for a given action creator which makes api calls. The tests still look very similar to the implementation (both are basically just configuration code), but we are now actually testing that these action creators are doing what we expect them to, without caring about how their implementation actually looks.


justinvdm:
@smn @Misto @projectmushroom depends #254, otherwise ready for review.

smn:
:+1:

PR #257, opened 2016-11-24
Assorted cleanup

justinvdm:
@smn @Mitso @projectmushroom ready for review

PR #253, opened 2016-11-23
No onComponentDidMount() for saving profile at end of onboarding

justinvdm:

PR #258, opened 2016-11-24
Use uids for finding view nodes in tests

justinvdm:


justinvdm:
@smn @Mitso @projectmushroom depends on #253, otherwise ready for review.

justinvdm:
Now depends on #257, merged in to avoid PR spaghetti causing confusing conflicts.

smn:
:+1:

PR #261, opened 2016-11-29
Clean up navigation

justinvdm:


src/__tests__/helpers-test.js
```diff
@@ -111,4 +112,15 @@ describe('helpers', () => {
...
   });
+
+  describe('pipeline', () => {
+    it('should thread the given value through a pipeline of functions', () => {
+      const fn = pipeline([
+        (a, b) => a + b,
+        (a, b) => a * b,
+      ]);
+
+      expect(fn(2, 3)).toEqual(15);
```
smn: Is this line a non-async deferred?

src/helpers.js
```diff
@@ -48,3 +48,10 @@ export const errorSink = (store, actions, fallback) => e => Promise.resolve(e)
   .then(switchError(actions))
   .then(store.dispatch)
   .catch(fallback);
+
+
+export const pipeline = fns => (v, ...args) => {
+  let res = v;
+  for (const fn of fns) res = fn(res, ...args);
+  return res;
+};
```
smn: not entirely sure but it feels like we're reimplementing Javascript's own Array.prototype.reduce function here? How is this different?

justinvdm: Ya, this could be written as `fns.reduce((res, fn) => fn(res, ...args), v)`, a bit simpler, will do that.

What would've been ideal would be to use lodash's [`flow()`](https://lodash.com/docs/4.17.2#flow), but it doesn't support extra arguments, which we need for passing the action argument when using this to create a pipeline of reducers.

justinvdm: I don't think I actually answered your question. This difference between this current implementation and one that uses `Array.prototype.reduce()` is that we're avoiding creating a functional call unnecessarily, which may/may not be important, depending on how often this function is called (most cases, probably not important).

I guess I've just been in the habit of avoiding creating functions unnecessarily if we don't know how a function may be used (which is the case for a helper function), but maybe here its overkill.

src/__tests__/helpers-test.js
justinvdm: `fn(2, 3)` is a normal, sync function call, and we're just comparing its result to an expected value. I think I might be missing your question though.

src/helpers.js
justinvdm: (sorry, edited reply ^, noticed mistakes)

justinvdm: done

justinvdm:
@smn ready for re-review

smn:
:+1:

PR #262, opened 2016-11-29
Poll for event updates

justinvdm:
We need a way to keep the event list on the journey screen up to date. We don't have a streaming-like option available here, so instead the plan is to poll for events every 2 minutes.


smn:
:+1:

PR #263, opened 2016-11-29
Scheduled call event view

justinvdm:
At the moment the view details are in the container, we should move these into a view so that the container is only providing the relevant data.


justinvdm:
screen shot 2016-11-29 at 5 31 43 pm
screen shot 2016-11-29 at 5 33 44 pm

smn:
:+1:

PR #264, opened 2016-11-30
Fix stepper navigation

justinvdm:
Recent work on `NavigationStack` and `Stepper` broke the things using them. Could've sworn I tested them out before submitting a PR but seems I did not.


justinvdm:
fixed-nav-call-notes
fixed-nav

smn:
:+1: bleh sorry, could've probably caught those if I was better at React & js

PR #267, opened 2016-12-01
Allow version suffixes when reading package.json version in build.gradle

justinvdm:
This will allow us to publish dev releases. The suffixes still can't be taken into account when calculating the version code google play uses, but provided we arent using suffixes outside of development releases, I think this is fine.

@smn ready for review


smn:
:+1:

PR #268, opened 2016-12-01
Document release process

justinvdm:
We are soon going to need to do dev releases (for the next feature to be deployed) and hotfix releases (for fixes to the currently live version of the app). `git flow release` can't help us for dev releases, since it merges changes from develop into master (we plan on using `master` as the stable branch for the live version of the app).

Since we need to diverge a bit from git flow, we should document how this new process will look.

@smn ready for review

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+git-tag-version=false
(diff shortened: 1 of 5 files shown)


justinvdm:
Needed to change process a bit to avoid merge conflicts when merging in hotfix releases into develop.

smn:
:+1:

PR #269, opened 2016-12-01
List chat messages

justinvdm:
@smn ready for review


justinvdm:
Not matching the designs yet, that is for later. Also, there are issues with the pattern background and nav tab ordering, but those are separate things I'll make separate PRs for.

screen shot 2016-12-01 at 10 56 44 pm

smn:
:+1:

PR #271, opened 2016-12-02
Fix PatternBackground when content is shorter than pattern image

justinvdm:
@smn ready for review


justinvdm:
screen shot 2016-12-02 at 11 40 09 am

PR #273, opened 2016-12-02
Add send message actions

justinvdm:
The plan is to maintain a set of pending messages (that are either sending or failed), alongside the set of sent and received messages. Then, when we get a success response, we clear the corresponding pending message.

This way all that needs to happen at the view layer is for these two sets to be merged before rendering.

This PR is for adding the actions needed for doing this. The logic for changing state based on these actions will happen in a different PR.


justinvdm:
@smn depends on #269, otherwise ready for review

smn:
:+1:

PR #274, opened 2016-12-05
Shape pending and sent message objects consistently

justinvdm:
Since we plan on merging the lists of pending and sent messages when providing the data to the view layer, we need the two types of messages to have the same property names.


justinvdm:
@smn ready for review

smn:
:+1:

PR #275, opened 2016-12-05
Reducer logic for message sending

justinvdm:
We need to clear the relevant pending message when the completed message is created via the api.

@smn depends on #274, otherwise ready for review


smn:
:+1:

PR #276, opened 2016-12-06
Message sending states

justinvdm:
Adds sending message states to view layer.

@smn ready for review for when you're feeling better


justinvdm:
screen shot 2016-12-06 at 10 06 33 pm

smn:
:+1: but good grief these React PRs are hard to review. I suppose a lot of this is just inherent to how react code is structured.

justinvdm:
@smn thanks for review

> 👍 but good grief these React PRs are hard to review. I suppose a lot of this is just inherent to how react code is structured.

howso? because of all the different modules and layers? If so, I think me moving stuff around probably didn't help in terms of how easy to read the diff was (sorry about that), but otherwise I found that once I was familiar with the different layers reviewing was actually felt a bit easier. It could well be that I've over-modularised the view components in this PR though, felt it simplified the code but probably did mean lots of scrolling around seeing how things are connected (sorry about that).

smn:
yeah, it wasn't a comment about your code but more about the number of files every change seems to involve and that does result in a fair amount of jumping around.

PR #277, opened 2016-12-07
Group messages

justinvdm:


justinvdm:
@miltontony || @smn (if you're feeling better): depends on #276, otherwise ready for review:

justinvdm:
screen shot 2016-12-07 at 12 15 44 pm

PR #280, opened 2016-12-07
Fetch recent events

justinvdm:
We poll for events every 2 minutes, and get back 100 items each time. Since it is unlikely for this many events to have appeared each time we poll, and since this has a chance of making the app slugish (by blocking the js thread while merging in the new items), it would be better to ask for fewer in this case.

@smn ready for review


miltontony:
👍

PR #281, opened 2016-12-07
Fetch recent messages

justinvdm:
Same as #280, but for messages.

@smn ready for review

Note: depends (implicitly) on #280.


miltontony:
👍

PR #282, opened 2016-12-07
Poll for recent messages

justinvdm:
The plan is to poll for the last 25 messages every 2 minutes. Usually we should be able to rely on new message firebase notifications for this, so this is a fallback for if and when firebase isn't pushing through notifications quickly enough. Fetching using the last `n` rather than fetching since the last time messages were fetched does mean we could potentially miss messages, but the latter approach is more work than we have time for at the mmoment, and a user sending more than 25 messages in 2 minutes seems like enough of an edge case.


justinvdm:
@smn depends on #281, otherwise ready for review

smn:
:+1:

PR #283, opened 2016-12-07
Add retry functionality for failed messages

justinvdm:

PR #284, opened 2016-12-08
Assorted fixes for messages

justinvdm:


justinvdm:
@smn depends on #283, otherwise ready for review.

smn:
:+1:

PR #285, opened 2016-12-08
Activity ordinals

justinvdm:
At the moment we rely on the ordering in which categories and activities are given by the api. This isn't the same as the ordering in which the items are arranged in wagtail

If no `ordinal` field is present, we rely on the ordering the api gives us when listing things.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+git-tag-version=false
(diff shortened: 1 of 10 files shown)


justinvdm:
Sorry for noise, based PR on wrong branch, recreated as #281.

PR #299, opened 2016-12-21
More robust profile field parsing

justinvdm:
Update profile api parsing to be more robust to non-empty string initial values for profile fields


codiebeulaine:
👍 wtih

PR #301, opened 2016-12-21
Use native picture picker

justinvdm:
RN's CameraRoll component appears to return full size images, which makes our current camera roll component not very stable.


justinvdm:

codiebeulaine:
👍 wtih ! Super cool how much code you replaced by using the native image stuff :)

PR #302, opened 2016-12-22
Schedule call time disappears when an activity is chosen

justinvdm:
This happens because the view re-renders, but no longer has the relevant time as state. Our options were to change ourn navigation stack to render all views, even if they aren't currently shown, hold onto the current schedule call state in redux while we are busy choosing an activity, or pass along the currently chosen date and time as context while choosing an activity. Option 3 seemed the most sane solution for what we have at the moment.

### DIFF
--- a/src/actions/__tests__/schedule-test.js
+++ b/src/actions/__tests__/schedule-test.js
@@ -12,6 +12,7 @@ import {
   addScheduledCall,
   addNextScheduledCall,
   openScheduledCall,
+  changeScheduledCallActivity,
   chooseScheduledCallCategory,
   chooseScheduledCallActivity,
   startScheduledCall,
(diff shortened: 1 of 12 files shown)


src/actions/__tests__/schedule-test.js
```diff
@@ -82,22 +83,40 @@ describe('schedule/actions', () => {
...
+            context: { foo: 21 },
+          },
         });
     });
   });
 
   describe('chooseScheduledCallActivity', () => {
-    it('should create an action for choosing a activity', () => {
-      expect(chooseScheduledCallActivity(23))
+    it('should create an action for choosing an activity', () => {
```
codiebeulaine: @justinvdm nice spelling change :)

codiebeulaine:
👍

PR #303, opened 2016-12-22
Fix keyboard disappearing after first character

justinvdm:
Turns out that this was a [reconciliation](https://facebook.github.io/react/docs/reconciliation.html) issue in `NavigationStack`. When a change happened somewhere, we would re-render the navigation stack. If the active route didn't change, it would render a single route rather than left and right "slides" to transition between. It seems that this change caused the react component to be re-mounted and rendered.

To fix this, we don't change the transition state if the active route has not changed.


codiebeulaine:
👍

PR #305, opened 2016-12-22
Functional device back button

justinvdm:
This PR adds the following behaviour when device back button is pressed:
  - if on landing screen, exit app
  - else if on a 'main' page (not overlays are showing), logout
  - else if not in a stepper, dismiss the current overlay
  - else if in a stepper (onboarding, call notes)
    - if not at step 0, go back a step
    - else revert to steps above


justinvdm:
Changed behaviour described in description to instead just exit the app if back is pressed on a 'main' screen.

codiebeulaine:
👍 if has been tested locally

PR #308, opened 2016-12-23
Only use multiline text input for steps that need it for Onboarding

justinvdm:


codiebeulaine:
:+1:

PR #309, opened 2016-12-23
Mark activity complete when a call note is added and its objective is achieved

justinvdm:


codiebeulaine:
👍

PR #310, opened 2016-12-23
Fix back button taps on main screen after login

justinvdm:
At the moment we ignore device back button taps for loading screens in the loading component. If the component is still being rendered even though it is not visible (as is the case for some navigation transitions), this will block back button taps.


codiebeulaine:
👍

PR #312, opened 2017-01-04
Parsing for rich text api fields

justinvdm:
To avoid duplicating code and doing more work than we need to, the plan is to draw components using [simple-markdown](https://github.com/Khan/simple-markdown) with the react-native-specific rules defined by [react-native-simple-markdown](https://github.com/CharlesMangwa/react-native-simple-markdown) (for both markdown-based rich text fields, and non-markdown based rich text fields defined directly in wagtail). This means that most of the time, we should only need to override styles, and the actual implementation of the rich text types is done for us already by react-native-simple-markdown. It also means we don't need different implementations for markdown-based rich text fields and non-markdown-based rich text fields.

First step for that ^ approach is to add parsing support for all rich text field types we get from wagtail, to allow us to parse these fields into a syntax tree recognised by [simple-markdown](https://github.com/Khan/simple-markdown).

@miltontony ready for review

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
     "react-native-image-picker": "^0.24.1",
     "react-redux": "^4.4.5",
     "redux": "^3.4.0",
-    "redux-thunk": "^2.0.1"
+    "redux-thunk": "^2.0.1",
+    "simple-markdown": "^0.2.1"
   },
   "devDependencies": {
     "babel-cli": "^6.7.5",
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "react-native-config": "^0.1.0",
     "react-native-fcm": "^2.0.3",
     "react-native-image-picker": "^0.24.1",
+    "react-native-simple-markdown": "^1.0.60-rc.3",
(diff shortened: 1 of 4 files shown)


justinvdm:

Mitso:
@justinvdm :100: 
Just throwing this out, there is Media and Page List tags from Wagtail, which are links to video or featured articles we don't use those or often but they exists.

justinvdm:
@Mitso thanks for letting me know :) not sure if those are accessible as wagtail stream field types (some of which I added support for in this PR), but will keep an eye on the field types we add on the wagtail side to make sure we have support for them on the app side.

justinvdm:
@nathanbegbie going to merge this so long, but would be great if you could review anyways, would give context for the PRs ahead.

nathanbegbie:
@justinvdm Agreed - busy reviewing at the moment, but my ES6 is super rusty, so will be a while before I get through this

justinvdm:
> @justinvdm Agreed - busy reviewing at the moment, but my ES6 is super rusty, so will be a while before I get through this

Thanks :) no rush

PR #313, opened 2017-01-05
Basics for RichText component

justinvdm:
This adds a `RichText` component that uses the syntax trees returned by `richText()` (adding in #312) as input.

See #312's description for some context on the approach taken for `RichText`'s implementation.

Styling to come in separate PRs.

### DIFF
--- a/src/components/RichText/__tests__/RichText-test.js
+++ b/src/components/RichText/__tests__/RichText-test.js
@@ -67,4 +67,29 @@ describe('RichText', () => {
       <RichText rules={rules}>{richText()('rule overrides')}</RichText>
     )).toMatchSnapshot();
   });
+
+  describe('shouldComponentUpdate', () => {
+    it('should compare props via a SameValueZero comparison', () => {
+      const content1 = richText()('foo');
+      const content2 = richText()('foo');
+
+      const props = {
+        styles: {},
+        rules: () => ({}),
+        children: content1,
+      };
+
+      const el = shallow(<RichText {...props} />);
+
+      expect(el.instance().shouldComponentUpdate({
+        ...props,
+        children: content1,
+      })).toEqual(false);
+
+      expect(el.instance().shouldComponentUpdate({
+        ...props,
+        children: content2,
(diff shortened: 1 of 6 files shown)


justinvdm:
screen shot 2017-01-05 at 1 05 31 pm

justinvdm:

Mitso:
@justinvdm +1 
Strikeout text isn't strike-through 

Not much to do with the review itself
I have tried to review with my basic understanding of React, react components and props etc.
I'm looking to understand a little better RichText Api parser of the Wagtail rich text fields
as well as overriding or the Markdown Rules, confused by the empty _src/components/RichText/rules.js_ component the styling makes sense.

justinvdm:
@Mitso thanks for review :)

> Strikeout text isn't strike-through
Yep, planning on fixing in separate PR

> confused by the empty src/components/RichText/rules.js 

This is just a placeholder for now. It will eventually include custom rules for cases where we want to define our own react components for a rich text type (we will be doing this for images).

A little bit of context that may help: To make things flexible, [simple-markdown](https://github.com/Khan/simple-markdown) allows one to specify a set of rules (each rule mapping to a rich text type, e.g. heading) to use to turn markdown text into react components.

One thing that can be defined for each rule is how to turn the markdown text into a "syntax tree" made up of nodes/elements/objects representing each heading, paragraph, image, group of text, etc. Another thing that can be defined for each rule that we care about more is turning a syntax tree node into a react component. We are relying on the [rules defined by react-native-simple-markdown](https://github.com/CharlesMangwa/react-native-simple-markdown/blob/master/rules.js), and just overriding the styling. If we want to define our own way of turning syntax tree node into a react component, we add it to `RichText/rules.js`.

`api/richText()` does the work of creating a syntax tree (either manually for wagtail fields that aren't markdown based, or automatically via simple-markdown for fields that are markdown based) from the data we get from the api. The `RichText` react component takes in the syntax tree as input and uses it to render everything (via react-native-simple-markdown's rules, and our own custom rules).

PR #316, opened 2017-01-06
Styling for rich text lists

justinvdm:

PR #317, opened 2017-01-06
Fixes for non-markdown rich text fields

justinvdm:
Now that we are able to draw these fields, there were a few minor differences between the syntax trees we create via our rich text components, and the ones created via simple-markdown.

@Mitso @nathanbegbie depends on #316, otherwise ready for review


justinvdm:
@Mitso @nathanbegbie ready for review now that #316 has landed.

nathanbegbie:
@justinvdm Looks good to me 👍

PR #318, opened 2017-01-06
Add rule for drawing rich text images

justinvdm:

### DIFF
--- a/src/components/RichText/Image.js
+++ b/src/components/RichText/Image.js
@@ -1,0 +1,26 @@
+import React, { PropTypes } from 'react';
+import { View, Image as RNImage } from 'react-native';
+
+import { ImageUrl } from 'src/api';
+
+
+const Image = ({
+  url,
+  styles = {},
+}) => (
+  <View style={styles.mentorshipImageContainer}>
+    <RNImage
+      style={styles.mentorshipImage}
+      source={url.resize(328, 183).toSource()}
+    />
+  </View>
+);
+
+
+Image.propTypes = {
+  styles: PropTypes.object,
+  url: PropTypes.instanceOf(ImageUrl).isRequired,
+};
+
+
+export default Image;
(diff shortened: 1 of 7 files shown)


justinvdm:
screen shot 2017-01-06 at 4 45 21 pm

src/components/RichText/Image.js
```diff
@@ -0,0 +1,26 @@
...
+
+
+const Image = ({
+  url,
+  styles = {},
+}) => (
+  <View style={styles.mentorshipImageContainer}>
+    <RNImage
+      style={styles.mentorshipImage}
+      source={url.resize(328, 183).toSource()}
```
nathanbegbie: @justinvdm Is there a particular reason that we're using these dimensions: `328, 183`?

justinvdm: Yep, its the dimensions given in the designs, so we specify these dimensions in the image's style declarations, and when telling the api at what size we want the image to be returned. React native requires us to [specify the image dimensions up front](https://facebook.github.io/react-native/docs/images.html#network-images) to prevent content jumping around when images load.

There are some subtleties here:
  -  units in react native are [device-independant](https://en.wikipedia.org/wiki/Device-independent_pixel), which means on a 3x density device, the units here would actually be `328`, `183`
  - if you look in the style declarations, you'll see that we max the image width at the device width, minus padding. So if the device's width is smaller than `328 x deviceScreenDensity`, we will instead resizing it to fit to this width, preserving aspect ratio
  - we preserve image dimension ratios, so if an image of a ratio different to these dimensions is being displayed:
    - "taller" ratios: after resizing to fit the width to 328, if the height is larger than 183, the image will be truncated
    - "wider" ratios: after resizing to fit the width, the height of the image will be scaled uniformly, but the actual space being taken up by the image will be kept at 183
    - images are hard :)

justinvdm: @nathanbegbie Fixed some mistakes in comment, please refresh before reading ^

justinvdm: Explaining this has made me realise we should contain taller images instead of truncating them, going to fix, thanks :)

justinvdm: Done

screen shot 2017-01-11 at 11 49 19 am

nathanbegbie:
+1

PR #319, opened 2017-01-06
Fetch recent events on call note create success

justinvdm:
At the moment we poll for new events to update the journey screen, and fetch recent events creating a new thing via the api if we know that an event would be created as a side effect. Call notes is one of those things, except we still need to set things up to fetch recent events when call notes are created.

### DIFF
--- a/src/effects.js
+++ b/src/effects.js
@@ -7,6 +7,7 @@ import { load } from 'src/actions/sync';
 import { sequence } from 'src/actionHelpers';
 import { setupNotifications } from 'src/actions/notifications';
 import { EVENT_POLL_TICK } from 'src/constants/events';
+import { CALL_NOTE_CREATE_SUCCESS } from 'src/constants/callNotes';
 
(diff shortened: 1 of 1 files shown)


justinvdm:
@Mitso @nathanbegbie reminder that this is ready for review

Mitso:
@justinvdm it's +1 from me.
Could be a superfluous question   
In what event  would the Call Notes be created a side effect and what point do you schedule a Call Note after creation ?

justinvdm:
@Mitso I think my description was a bit confusing. "Events" are things that are created on the backend when some things happen in the application (e.g. notes were filled in for a call, a call was scheduled, etc). So for this case, an event is created as a side effect on the backend when we talk to the backend to create a call note. We know that the event has been created, but we haven't yet fetched it from the backend, so `listRecentEvent` fetches it from the backend (it should really be named `fetchRecentEvent`). Does that help?

PR #320, opened 2017-01-06
Fix rich text strikethrough

justinvdm:

### DIFF
--- a/src/components/RichText/__tests__/RichText-test.js
+++ b/src/components/RichText/__tests__/RichText-test.js
@@ -26,7 +26,8 @@ text
 
 [link](http://www.google.com)
 
-*italics*
+*italics via asterisks*
+_italics via underscores_
 **bold**
 ***bold italics***
 ~~strikeout~~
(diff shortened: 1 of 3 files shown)


justinvdm:
screen shot 2017-01-06 at 5 49 32 pm

src/components/RichText/__tests__/RichText-test.js
```diff
@@ -10,34 +10,34 @@ describe('RichText', () => {
...
+text
 
-        [link](http://www.google.com)
+[link](http://www.google.com)
 
-        *italics*
-        **bold**
-        ***bold italics***
-        ~~strikeout~~
+*italics*
```
nathanbegbie: @justinvdm Will italics and bold always be denoted by an asterisk (*) and not underscore (_) ? Is there a particular version/style of markdown that we're using?

nathanbegbie: Ah, I see underscores are used for underlining in #321

justinvdm: We’re relying on https://github.com/Khan/simple-markdown which supports both, added test for `_` case.

justinvdm: screen shot 2017-01-11 at 12 17 14 pm

nathanbegbie:
+1

PR #321, opened 2017-01-06
Support rich text underlining

justinvdm:

### DIFF
--- a/src/components/RichText/rules.js
+++ b/src/components/RichText/rules.js
@@ -1,7 +1,22 @@
+import { isString } from 'lodash';
 import React from 'react';
-import { Text } from 'react-native';
+import { View, Text } from 'react-native';
 
 import Image from './Image';
+
+
+const reactTextType = style => (node, output, state) => {
+  const children = isString(node.content)
+    ? node.content
+    : output(node.content, {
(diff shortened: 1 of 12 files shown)


justinvdm:
screen shot 2017-01-06 at 6 06 45 pm

justinvdm:
(though it does depend on #320 and #318)

src/components/RichText/rules.js
```diff
@@ -1,4 +1,27 @@
...
+import Image from './Image';
+
+
+const rules = styles => ({
+  mentorshipImage: {
+    react: node => (
+      <Image styles={styles} url={node.url} />
+    ),
+  },
+  u: {
```
nathanbegbie: Maybe be more explicit and rename to underline?

justinvdm: We're kind of limited to the names defined by `simple-markdown`, for e.g. for [`u`](https://github.com/Khan/simple-markdown/blob/master/simple-markdown.js#L1224-L1241), so that it can map the syntax tree's node type field (in this case `u`) to a corresponding rule.

nathanbegbie:
+1 from me, once conflicts resolved

PR #324, opened 2017-01-10
Support code in rich text

justinvdm:
Very unlikely this will be used, but at the moment it just breaks. We should at least show something, rather than have the app break.

Inline code obviously needs some padding, but [this isn't possible](https://github.com/facebook/react-native/issues/3233) for react native `Text` yet, and we don't have the luxury of being able to wrap code in a `View` for this case, since the tree for inline code ends up being nested inside a `Text` element (`View` elements cannot be nested inside `Text`s, at least not yet).

### DIFF
--- a/src/components/RichText/__tests__/RichText-test.js
+++ b/src/components/RichText/__tests__/RichText-test.js
@@ -5,7 +5,6 @@ import { richText } from 'src/api';
 
 describe('RichText', () => {
   it('should render typography', () => {
-    // TODO get code blocks working
     expect(render(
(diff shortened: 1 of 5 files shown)


justinvdm:
screen shot 2017-01-10 at 12 03 42 pm

justinvdm:
@nathanbegbie @Mitso depends on #321, otherwise ready for review.

nathanbegbie:
Pending the other PRs, +1 from me - although you may want @Mitso to also take a look

PR #325, opened 2017-01-11
Navigation stack throws away old routes' view state

justinvdm:
At the moment, `NavigationStack` (the component responsible for transitioning between routes) only renders the view if it is being transitioned to or from. This means that we lose view state for all other routes on the stack that are not being shown.

To fix this, we should still render the hidden views, except not actually display them.

### DIFF
--- a/src/components/NavigationStack/index.js
+++ b/src/components/NavigationStack/index.js
@@ -174,7 +174,8 @@ class NavigationStack extends Component {
       <View
         key={route.key}
         uid={route.key}
-        style={[styles.slide, isHidden && styles.slideIsHidden]}>
+        style={[styles.slide, isHidden && styles.slideIsHidden]}
+      >
         {this.renderRoute(route)}
       </View>
     );
--- a/src/components/NavigationStack/index.js
+++ b/src/components/NavigationStack/index.js
@@ -162,7 +162,7 @@ class NavigationStack extends Component {
     const obj = this.props.routes[key];
 
     if (React.isValidElement(obj)) {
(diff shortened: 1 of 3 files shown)


src/components/NavigationStack/index.js
```diff
@@ -96,6 +100,57 @@ class NavigationStack extends Component {
...
+      default:
+        res = [curr];
+        break;
+    }
+
+    return res.map(getCurrent);
+  }
+
+  getRoutesExcluding(excludes) {
+    const keys = excludes.map(({ key }) => key);
```
Mitso: Since _excludes.map(({key})_ loops through an the object what is the structure of_ => key_ )?

justinvdm: @Mitso not quite understanding, could you elaborate?

Mitso: Oh I mean like what's returned by const keys?

justinvdm:
@Mitso thanks for review :) +1 on screen recording. In the one below, when adding an activity to the scheduled call, the scheduled call screen is hidden (the other screens would not render correctly if it wasn't being hidden properly):

navigation-stack-hidden-views

src/components/NavigationStack/index.js
justinvdm: Ah :) `excludes` would look something like `[{ key: 'ROUTE_A' }, { key: 'ROUTE_B' }]`, so `keys` would look like `['ROUTE_A', 'ROUTE_B']`.

PR #326, opened 2017-01-11
"Three words" onboarding screen allows empty strings as words

justinvdm:
For e.g. `foo,bar,` would count as words `['foo', 'bar', '']`, when it should be `['foo', 'bar']`.


nathanbegbie:
+1

PR #329, opened 2017-01-12
Fix text wrapping for call note detail view

justinvdm:


justinvdm:
Made against incorrect branch, closing. Sorry for the noise.

PR #330, opened 2017-01-12
Fix text wrapping for call note detail view

justinvdm:
At the moment, if the content for "discussion with activity" section is long enough, it trails off the screen instead of wrapping.

Ended up including some adjustment and alignment fixes for the same screen, which had some implications for other screens using the same styles.


justinvdm:
@projectmushroom @Mitso @nathanbegbie ready for review

justinvdm:
screen shot 2017-01-12 at 4 15 54 pm
screen shot 2017-01-12 at 4 44 48 pm
screen shot 2017-01-12 at 4 40 21 pm

PR #331, opened 2017-01-12
Require words to have at lease one letter for "three words" onboarding screen

justinvdm:


justinvdm:
@projectmushroom @Mitso @nathanbegbie ready for review

PR #327, opened 2017-01-11
Boilerplate for blogs

justinvdm:
@Mitso @nathanbegbie ready for review

PR #332, opened 2017-01-13
Parse blog post body content using rich text components

justinvdm:


justinvdm:
@nathanbegbie @Mitso depends on #327, otherwise ready for review

PR #334, opened 2017-02-20
Stub TouchableNativeFeedback in tests

justinvdm:
Element trees rendered by `TouchableNativeFeedback` are currently hidden from our snapshot tests, instead being replaced by this message:

> TouchableNativeFeedback is not supported on this platform!


justinvdm:
@codiebeulaine, @Mitso, @nathanbegbie ready for review.

Only the two first files are important. The remaining diffs of the `.snap` files are just the updated snapshots after fixing things.

PR #335, opened 2017-02-20
Use content prop instead of children prop for RichText content

justinvdm:
I've recently discovered that objects aren't allowed us children.


justinvdm:
@Mitso, @nathanbegbie, @codiebeulaine depends on #328, otherwise ready for review

PR #338, opened 2017-03-31
Display scrollbar on call note reflections input when when available height exceeded

justinvdm:


codiebeulaine:
👍

PR #339, opened 2017-04-03
Hide blog nav tab

justinvdm:
Blogs was the next in-development feature. We are putting it on hold because of a change in project direction. We still want to use the `develop` branch for in-development features though, so the plan is to just not surface this feature in the UI for now.


justinvdm:
@nathanbegbie could you review this?

nathanbegbie:
@justinvdm :+1:

PR #341, opened 2017-04-06
Add dev entry point

justinvdm:
When building a view it is often useful to render it in isolation rather than navigate to it after every reload. We've been modifying `index.android.js` instead, but this is problematic: it is easy to accidentally commit changes to this file made for testing things in development, for example.

This PR makes that possible by adding a config option. If `USE_DEV_APP` is set to `true` (which can be configured in `src/config/dev.js`), and we are in development mode (`__DEV__ === true`), this will use `src/app.dev.js` as the entry point for the app instead of `src/app.js`.

If one just wanted to add some side effects for testing something in development, this would also be possible: `src/app.dev` would simply need to export `src/app.js`'s default as its default.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,7 @@ android/app/google-services.json
 .env.qa
 .env.prd
 src/config/dev.js
+src/app.dev.js
 
 *.keystore
 secrets.tar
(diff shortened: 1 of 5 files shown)


justinvdm:

nathanbegbie:
@justinvdm  :+1:

PR #342, opened 2017-04-12
Don't show hidden categories and activities

justinvdm:
We show hidden categories and activities when adding an activity to a scheduled call.


nathanbegbie:
+1

PR #345, opened 2017-05-03
Add a ControlledStepper component

justinvdm:
At the moment we have a stepper that does the navigation between steps itself. This isn't enough for call notes v2, where there is more complex logic dictating which step one sees next.

We need a new component that just takes in navigation state and draws the right thing, so that we can do the navigation logic for more complex use cases like this inside a reducer.


nathanbegbie:
:+1:

PR #346, opened 2017-05-05
Support function-based routes prop in NavigationStack

justinvdm:
This gives us a less awkward api for cases where we don't have existing elements we want to define routes for.


src/components/NavigationStack/__tests__/NavigationStack-test.js
```diff
@@ -192,4 +192,23 @@ describe('NavigationStack', () => {
     expect(el.find('A').length).toEqual(1);
     expect(el.find('B').length).toEqual(1);
   });
+
+  it('should function-based routes prop', () => {
```
nathanbegbie: I think there's a missing word "support" here.

nathanbegbie:
:+1:

PR #347, opened 2017-05-05
Feature/add new call notes assets

nathanbegbie:


nathanbegbie:
@justinvdm ready for review. Are there any assets that I've missed?

src/constants/images.js
```diff
@@ -66,6 +66,9 @@ import EVENT_MENTEE_BORED from 'app/assets/bored-journey.png';
 import EVENT_MENTEE_CONFUSED from 'app/assets/confused-journey.png';
 import EVENT_MENTEE_UPSET from 'app/assets/upset-journey.png';
 import EVENT_MENTEE_WITHDRAWN from 'app/assets/withdrawn-journey.png';
+import EVENT_MENTEE_EXCITED from 'app/assets/excited-journey.png';
+import EVENT_MENTEE_ENGAGED from 'app/assets/engaged-journey.png';
```
justinvdm: I think these are just for the events on the journey screen, we will need assets for the larger versions of the same image (see the code block above this one).

justinvdm:
@nathanbegbie should we update this now that we have the new assets?

nathanbegbie:
@justinvdm readey for re-review

justinvdm:
:+1:wtih

nathanbegbie:
@justinvdm Travis was not happy. 
Are you okay with how I'm now combining objects in `constants/callNotes` ?

justinvdm:
@nathanbegbie hmm, wondering if it is "too" DRY, they feel like mappings we want to maintain separately (for e.g, if v2 mood images ended up _not_ using some v1 images). Not too sure though, and unimportant enough that I think this should land as is.

PR #350, opened 2017-05-08
Call notes v2 boilerplate

justinvdm:
@nathanebegbie ready for review


src/views/CallNoteStepsV2/ObjectiveAchieved/index.js
```diff
@@ -0,0 +1,24 @@
+import React, { PropTypes } from 'react';
+
+import { FormStep } from 'src/components';
+
+
+const ObjectiveAchieved = ({
+  callNote,
+  onChange,
+  ...props,
+}) => callNote && onChange && (
```
nathanbegbie: Why do we have this logic here? Could `propTypes` not catch this?

nathanbegbie:
:+1:

PR #351, opened 2017-05-08
Basic navigation for call note v2

justinvdm:

PR #354, opened 2017-05-11
CallResult call note v2 step and related changes

justinvdm:

The first of the steps. Involved quite a few changes to base components, but hopefully this clears the path for the other steps a bit.


justinvdm:
screen shot 2017-05-11 at 11 50 15 am
screen shot 2017-05-10 at 2 23 34 pm

justinvdm:
@nathanbegbie sorry forgot to say, this one depends on #351 and #353.

src/reducers/callNotes.js
```diff
@@ -6,64 +7,45 @@ import { AUTH_LOGOUT } from 'src/constants/auth';
...
 
-const callNote = (state = {}, action) => {
+export default (state, action) => {
   switch (action.type) {
     case AUTH_LOGOUT:
+      return createInitialState();
+
     case constants.CALL_NOTE_CREATE_OPEN:
+      return merge(createInitialState(), {
+        metadata: {
```
nathanbegbie: @justinvdm why use `merge` here instead of the usual object with `...` spread?

src/reducers/callNotes.js
```diff
@@ -6,64 +7,45 @@ import { AUTH_LOGOUT } from 'src/constants/auth';
...
 
 
-const callNote = (state = {}, action) => {
+export default (state, action) => {
   switch (action.type) {
     case AUTH_LOGOUT:
+      return createInitialState();
+
     case constants.CALL_NOTE_CREATE_OPEN:
+      return merge(createInitialState(), {
```
nathanbegbie: @justinvdm why use `merge` here instead of the usual object with `...` spread?

justinvdm: [`merge`](https://lodash.com/docs/4.17.4#merge) does a "deep" merge while `...` allows for a shallow merge. We could get away with `...` at the moment since `metadata` only has a single property, but to be more future proof to `metadata` getting other properties, we would need to `...` twice:

```js
const state = createInitialState();

return {
  ...state,
  metadata: {
    ...state.metadata,
    actionType: constants.ADD_IMMEDIATE,
  },
};
```

`merge` simplifies this a bit

nathanbegbie:
:+1: I'll merge once travis is happy

PR #357, opened 2017-05-15
Call note v2 mood step

justinvdm:


justinvdm:
screen shot 2017-05-15 at 7 28 58 pm
screen shot 2017-05-15 at 7 29 38 pm
screen shot 2017-05-15 at 7 30 35 pm

nathanbegbie:
@justinvdm Failing lint test here - I'm also a little bit concerned that on smaller screens it's not obvious that the screen is scrollable, but I'm not sure we have the time to solve that.

Could you also please explain the use of the `Dummy` component?

justinvdm:
@nathanbegbie thanks for review

re linting error - fixed.

> I'm also a little bit concerned that on smaller screens it's not obvious that the screen is scrollable

Agreed, though it sounds like it is a thing to look into in a second iteration rather.

> Could you also please explain the use of the Dummy component?

In the tests for the new `defineRadio` higher-order component, we need to inspect the props we instantiate the component with. For this, any component would do, so we could use, for e.g. `View`, except that we would get a validation warning since we would be given it props it is not expecting. We basically need a component that receives arbitrary props and doesnt really do anything with them so we can use it to inspect the props.

nathanbegbie:
:+1:

PR #359, opened 2017-05-16
Call note v2 steps: change activity

justinvdm:

This PR will hopefully make more sense once the corresponding backend changes have been made.


src/constants/callNotes.js
```diff
@@ -26,14 +26,16 @@ export const CALL_NOTES_MENTEE_ENGAGED = 'engaged';
...
 export const CALL_NOTES_VIEW_ALL = 'CALL_NOTES_VIEW_ALL';
-
 export const CALL_NOTE_CHOOSE = 'CALL_NOTE_CHOOSE';
 
 export const ADD_RETROACTIVELY = 'ADD_RETROACTIVELY';
 export const ADD_IMMEDIATE = 'ADD_IMMEDIATE';
 
+export const CALL_NOTE_ACTIVITY_CHANGE = 'CALL_NOTE_ACTIVITY_CHANGE';
+export const CALL_NOTE_CATEGORY_CHOOSE = 'CALL_NOTE_CATEGORY_CHOOSE';
+export const CALL_NOTE_ACTIVITY_CHOOSE = 'CALL_NOTE_ACTIVITY_CHOOSE';
```
justinvdm: @nathanbegbie what I forgot in the conversation we had the other day about naming of constants is that we do in fact need them namespaced, otherwise the constant's values themselves will collide in reducers. For e.g, if we called this one `ACTIVITY_CHOOSE`, it would collide with the same constant in `constants/activities.js`, and since we looking at the actual string values when switching in the reducer, these values are would be the same even though they are from different modules.

src/api/schemas.js
```diff
@@ -12,3 +12,8 @@ export const CallNote = new Schema('callNotes');
 export const Message = new Schema('messages');
 export const PendingMessage = new Schema('pendingMessages');
 export const BlogPost = new Schema('blogPosts');
+
+
+CallNote.define({
+  callData: Call,
+});
```
nathanbegbie: @justinvdm I just want to make sure I'm understanding this.

We're using normalizr to structure data returned from and sent to the API. 
We add the call schema to the CallNote, because in one of our use cases, a person may change the activity of a call. This means that we need to include those changes to `call` as well as `callNote` - so we use a single call to a particular endpoint, with both sets of data encapsulated, which can then be processed by the backend.

src/reducers/__tests__/callNotes-test.js
```diff
@@ -106,4 +106,24 @@ describe('reducers/callNotes', () => {
         }));
     });
   });
+
+  describe('CALL_NOTE_ACTIVITY_CHOOSE', () => {
+    it('should reset the activity progress', () => {
+      const { callNote } = reduce(void 0, actions.chooseCallNoteActivity(23));
```
nathanbegbie: I think if we're going to switch to using `undefined` instead of `void 0` in some places, we may as well do it here.

src/api/schemas.js
justinvdm: 🍰  spot on

src/reducers/__tests__/callNotes-test.js
justinvdm: How about we do the global replace of all `void 0`s once all the currently open PRs land?

src/api/schemas.js
justinvdm: this is important, since we don't want only half an update to succeed if connection is flaky

src/reducers/__tests__/callNotes-test.js
nathanbegbie: +1 on that

scripts/helpers.js
```diff
@@ -34,6 +34,8 @@ export const mock = () => {
 };
 
 
+export const propEquals = (name, id) => node => node.prop(name) === id;
```
nathanbegbie: Nice!

nathanbegbie:
:cake: 👍

scripts/helpers.js
nathanbegbie: Although I'm just wondering whether the args should be called `propName` and `propValue`

nathanbegbie: No wait, that doesn't make sense. Maybe `desriredPropName` instead of `propValue`? I dunno. #bikeshedding

justinvdm: woops, `id` was a mistake `value` makes sense. `propName` and `propValue` is maybe overboard if `propEquals` makes it clear enough that we are talking about a property name and value.

PR #362, opened 2017-05-18
Debounce form steps

justinvdm:
This prevents users from moving further ahead in the stepper than they should because of tapping the next button multiple times.

@nathanbebgie ready for review

### DIFF
--- a/src/constants/behavior.js
+++ b/src/constants/behavior.js
@@ -1,0 +1,1 @@
+export const PAGING_DEBOUNCE_INTERVAL = 250;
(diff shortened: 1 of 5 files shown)


src/constants/behavior.js
```diff
@@ -0,0 +1 @@
+export const PAGING_DEBOUNCE_INTERVAL = 250;
```
nathanbegbie: This might be a silly question, but why do we have an interval at all? Do we not want to only call the function once - in which case use something other than debounce? Or is this akin to a global variable - so that if the user returned to that screen they wouldn't be able to press the button again?

justinvdm: Good question, its because we want the button to still be press-able if the user returned to the screen with those buttons.

nathanbegbie:
:+1:

PR #358, opened 2017-05-16
Feature/gemmentor 439 password management

nathanbegbie:

PR #365, opened 2017-05-22
Feature/implement status message forgot password

nathanbegbie:


nathanbegbie:
@justinvdm ready for review - only thing I'm wondering is if I need to create tests for different states in the `ForgotPasswordReset` view?

nathanbegbie:
screen shot 2017-05-22 at 11 39 43 am
screen shot 2017-05-22 at 11 36 53 am

justinvdm:
@nathanbegbie could I wait for #358 to land first before reviewing?

nathanbegbie:
@justinvdm ready for review

src/components/StatusMessage/index.js
```diff
@@ -0,0 +1,29 @@
...
+}) => (
+  <Text
+    style={[
+      baseStyles.message,
+      theme.message,
+    ]}
+  >
+    {message}
+  </Text>
+);
```
justinvdm: If this component is simply different styling for text (no different markup), the pattern we have been using is to add a new type or theme for the text component in `src/components/Text/styles.js`, and simply provide that type or theme when using the text, e.g. `<Text style={Text.types.textInputMessage}>`.

Is there a reason I may be missing for why we need a separate component?

Also, if this change is going to be more work than it is worth, happy to stay with this approach.

src/containers/ForgotPasswordResetContainer.js
```diff
@@ -4,7 +4,9 @@ import ForgotPasswordReset from 'src/views/ForgotPasswordReset';
 import { dismissScreen } from 'src/actions/navigation';
 import { resetForgotPassword } from 'src/actions/forgotPassword';
 
-export const mapStateToProps = state => state.forgotPassword;
+export const mapStateToProps = state => {
+  return { status: state.forgotPassword.status.type };
```
justinvdm: The `status` idea is meant to be a data structure encapsulating the type of a status, and any possible extra details necessary for displaying that status (e.g. for a validation error status, the details could be validation errors).

With that in mind, I think it makes sense to take in the entire status object and `.type` on it in the view, for consistency with the status objects that do have details.

src/components/StatusMessage/index.js
nathanbegbie: I think I was thinking of using it as a way of passing in the display logic, as opposed to internal logic ala `LoginStatusMessage`. But it turned out that it doesn't need to display different messages and it's better to expose that logic in the view anyway.

I'd rather follow convention and make the changes - shouldn't take too long.

src/views/ForgotPasswordReset/index.js
```diff
@@ -22,12 +37,21 @@ class ForgotPasswordReset extends React.Component {
...
+  )
+
   inputIsValid = () => (
     !((this.state.token !== '') &&
       (this.state.newPassword !== '') &&
       (this.state.newPassword === this.state.checkNewPassword))
   )
 
+  canPressButton = () => (
+    this.inputIsValid &&
```
justinvdm: should this perhaps be `inputIsValid()`, and is there possibly a false positive test we need to fix for this?

src/views/ForgotPasswordReset/index.js
```diff
@@ -72,15 +110,20 @@ class ForgotPasswordReset extends React.Component {
           value={this.state.checkNewPassword}
           onChangeText={this.setCheckNewPassword}
         />
-
-        {/* TODO: <StatusMessage {...status} /> */}
+        {this.displayPasswordErrorMessage() &&
+          <StatusMessage
+            message={'Passwords don\'t match'}
```
justinvdm: I think we should use the `children` prop in `StatusMessage` so we can have this be `<StatusMessage theme={...}>Passwords don't match</StatusMessage>`. That is if we don't go the `<Text>` route I mentioned above though.

src/views/ForgotPasswordReset/index.js
```diff
@@ -39,6 +63,13 @@ class ForgotPasswordReset extends React.Component {
     });
   }
 
+  displayPasswordErrorMessage = () => {
+    return (
+      this.state.newPassword !== this.state.checkNewPassword
+      && this.state.newPassword !== ''
+    );
```
justinvdm: Would this not mean that we would display the error message if `newPassword` was non-empty by `checkNewPassword` was empty? I think we should only display an error if both aren't empty.

src/views/ForgotPasswordReset/index.js
justinvdm: Ah, do we use `canPressButton`?

justinvdm:
Left some minor comments, looks good :) I think we still need tests for the different states that `ForgotPasswordReset` view can be in.

src/containers/ForgotPasswordResetContainer.js
nathanbegbie: So it would just be 
```
return state.forgotPassword.status;
```
?

justinvdm: yep

src/views/ForgotPasswordReset/index.js
nathanbegbie: Yes, sorry didn't update the function. it's used by the `disabled` prop with the button

nathanbegbie:
@justinvdm ready for re-review

src/components/Text/styles.js
```diff
@@ -45,6 +45,16 @@ export const types = StyleSheet.create({
     fontFamily: FONT.BOLD,
     fontWeight: FONT_WEIGHT.BOLD,
   },
+
+  textInputErrorMessage: {
+    color: colors.STATUS_MESSAGE_TEXT_ERROR,
+    fontSize: 14,
+    textAlign: 'left',
+    marginTop: -20,
```
justinvdm: Oh, sorry for not noticing this earlier. A little worried about this, it means we are implicitly depending on the element on top of this element having a bottom margin or padding that is `20` or more. Wondering if we could instead change that element in the place that this is used to have less of a margin?

src/views/ForgotPasswordEmail/index.js
```diff
@@ -70,10 +69,9 @@ class ForgotPasswordEmail extends React.Component {
         />
         {
           this.props.status === FORGOT_PASSWORD_EMAIL_STATUS_BAD_ADDRESS &&
-          <StatusMessage
-            message={'Not a valid email address'}
-            theme={StatusMessage.themes.textInputMessage}
-          />
+          <Text style={Text.types.textInputErrorMessage}>
+            {'Not a valid email address'}
```
justinvdm: We can ditch the `{'` now I think :)

src/containers/ForgotPasswordEmailContainer.js
```diff
@@ -5,7 +5,8 @@ import { dismissScreen } from 'src/actions/navigation';
 import { emailForgotPasswordToken } from 'src/actions/forgotPassword';
 
 export const mapStateToProps = state => {
-  return state.forgotPassword.status;
+  console.log(state.forgotPassword.status);
```
justinvdm: Stray debugging line I think

src/components/Text/styles.js
nathanbegbie: I did wonder about that - but this style type is specifically for TextInputs, which have `margin: 20` by default, so it allows us to override this behaviour without having to constantly override the textinput that it provides info on.
Er, basically I think implicit is good here.

justinvdm: Ah, sorry missed that, +1

src/views/ForgotPasswordReset/index.js
```diff
@@ -1,18 +1,34 @@
...
   }
 
   constructor(props) {
     super(props);
     this.state = {
       token: '',
-      newPassword: '',
-      checkNewPassword: '',
+      newPassword: this.props.newPassword || '',
+      checkNewPassword: this.props.checkNewPassword || '',
```
justinvdm: I think we should call these `initialNewPassword` and `initialCheckNewPassword` to make it clear that these are only seed values, and that we don't care about their values changing later on since the component has already been initialised (otherwise its the duplicate source of truth thing that there used to be react docs for, though it seems to have disappeared, there is [this](https://medium.com/@justintulk/react-anti-patterns-props-in-initial-state-28687846cc2e) though).

src/views/ForgotPasswordReset/__tests__/ForgotPasswordReset-test.js
```diff
@@ -18,6 +25,61 @@ describe('ForgotPasswordReset', () => {
...
+    const el = createComponent({
+      status,
+    });
+
+    expect(render(el)).toMatchSnapshot();
+  });
+
+
+  it(('should display the password mismatch error message '
+      + 'when passwords are mismatched'), () => {
```
justinvdm: Wish there was a better way for us to do this but I guess there isn't.

justinvdm: Erm, long test descriptions, that is.

src/views/ForgotPasswordReset/__tests__/ForgotPasswordReset-test.js
```diff
@@ -18,6 +25,61 @@ describe('ForgotPasswordReset', () => {
...
+    const el = render(createComponent({
+      newPassword: 'foo',
+      checkNewPassword: 'foo',
```
justinvdm: These passwords don't mismatch but the test description suggests that is what we are testing.

nathanbegbie: Whoops! Good catch

src/views/ForgotPasswordReset/__tests__/ForgotPasswordReset-test.js
nathanbegbie: Yeah - I'm sure there's a better way if we had time . . . we can put it on the wishlist

justinvdm:
Looks good, left a few more minor comments :) :+1: once they are resolved.

PR #367, opened 2017-05-23
Call note v2 fixes

justinvdm:

### DIFF
--- a/src/containers/ActivityContainer.js
+++ b/src/containers/ActivityContainer.js
@@ -1,4 +1,4 @@
-import { filter, last } from 'lodash';
+import { findLast } from 'lodash';
 import { connect } from 'react-redux';
 import Activity from 'src/views/Activity';
 
@@ -8,7 +8,7 @@ import {
   getCategory,
   getActivity,
   getNextScheduledCall,
-  getCallNotes,
+  getCallsWithCallNotes,
 } from 'src/store/helpers';
 
 import {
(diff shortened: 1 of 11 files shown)


justinvdm:
screen shot 2017-05-23 at 3 18 58 pm
screen shot 2017-05-23 at 3 18 26 pm
screen shot 2017-05-23 at 3 16 49 pm

src/containers/ActivityContainer.js
```diff
@@ -18,18 +18,25 @@ import {
 } from 'src/actions/activities';
 
 
+const getLatestCallNote = (state, activityId) => {
```
nathanbegbie: It might be better convention to put this in `store/helpers.js` ?

nathanbegbie: Also better for testing purposes

src/views/Activity/index.js
```diff
@@ -164,16 +164,16 @@ const Status = ({
...
   onSchedulePress,
   onReschedulePress,
   onViewCallNotesPress,
 }) => {
-  if (latestCallNotes) {
+  if (latestCall) {
     return (
       <ActionButton style={style} onPress={() => onViewCallNotesPress(id)}>
-        View call notes
+        Add or view notes
```
nathanbegbie: I'm just wondering if this warrants renaming of the action - because the code does not suggest that this would enable the user to add call notes given that the associated action is `onViewCallNotesPress` which is bound to `viewActivityCallNotes`.

nathanbegbie:
One comment about naming, otherwise +1

src/views/Activity/index.js
justinvdm: The action name is maybe misleading, it is meant as "view the call notes list screen" (where adding/viewing is possible). If the actions on that screen were to change (for e.g. editing), then the name would get even more awkward `onViewAddOrEditCallNotesPress`. `onViewCallNoteListPress` is probably better, but I think in the app in general we haven't been referring to it as the "call notes list", so I think renaming this should ideally also involve renaming a bunch of other things, which is probably more work than its worth at this point.

Happy to just rename this one to `onViewCallNotesListPress` (or a different name if you think of something else) though, if you think it makes sense.

nathanbegbie: Ah yes - I missed that it was Note*s*
Hmmm. Yeah I think `onViewCallNotesListPress` makes it more obvious where it's going.

justinvdm:
@nathanbegbie +1 on https://github.com/praekeltfoundation/mentorship-droid/pull/367/commits/cf831a9ae0d72c8aa3be983f9c1e4d87f6fa01ef?

nathanbegbie:
:+1:

PR #369, opened 2017-05-24
Fix navigation stack to still render even if routes hasnt changed

justinvdm:


nathanbegbie:
:+1:

PR #371, opened 2017-05-25
Parse and use activityData property returned from call note api methods

justinvdm:


nathanbegbie:
:+1:

